PR #8, opened 2019-12-25
Move over to icarus for deployment

davish:
Took this mostly from basics


k8s/values.yaml
```diff
@@ -0,0 +1,9 @@
+deploy_version: 0.1.0
+image_tag: latest
+
+applicstions:
+  - name: student-life
+    image: pennlabs/student-life
```
pawalt: Add required `extraEnv` variables. See https://github.com/pennlabs/student-life/blob/master/k8s/chart/templates/deployment.yaml#L26 for which variables to add.

PR #9, opened 2020-01-11
Apply template

ArmaanT:
This PR applies the latest version of template-django which includes the necessary config to update to the latest version of Icarus. Additionally, Platform OAuth credentials and a sentry project have been set up.


studentlife/settings/base.py
```diff
@@ -1,5 +1,5 @@
 """
-Django settings for pennlabs project.
+Django settings for studentlift.
```
davish: Typo

studentlife/settings/production.py
```diff
@@ -1,43 +1,32 @@
...
+import sentry_sdk
+from sentry_sdk.integrations.django import DjangoIntegration
 
-from studentlife.settings.base import *  # noqa
+from studentlife.settings.base import *  # noqa: F401, F403
+from studentlife.settings.base import BACKEND_DOMAIN, DATABASES
 
 
-DEBUG = True
+DEBUG = False
```
davish: For now we should keep debug to true, since this isn’t a prod app yet and we’ll be changing all our credentials before it goes into production. Just in case iOS has more issues it’s easier to debug.

pawalt:
Make sure to fix uwsgi setttings

studentlife/settings/base.py
ArmaanT: Fixed

studentlife/settings/production.py
ArmaanT: Makes sense

PR #10, opened 2020-01-26
Fixed CSRF by extending SessionAuthentication

r3khaan:
Simply overrode the enforce_csrf method.


r3khaan:
Failing some of the linting tests, honestly don't know what they mean

ArmaanT:
I know it's a bit late, but since we're not using basic auth on any route why don't we set DRF default `authentication_classes=[]`? That way we don't need to provide a custom authentication class or anything?

PR #14, opened 2020-02-21
Dining Balances

astrike30:
I created a branch for dining endpoints.

Added:
- Dining Django app dining endpoints.
- Basic Balance information for a user (currently hardcoded to my pennid for testing).
- `django-runtime-options` dependency to store the end and beginning of semester date.
- `utils.py` file to main student-life app with functions to get a new end and beginning of semester date. This uses code taken from PennSDK which is being phased out and hits a Penn calendar API.

Sample JSON for what is returned at the `/dining` endpoint as of now:

```json
{
   "swipes":108,
   "dining-dollars":20,
   "guest-swipes":5,
   "start-of-semester":"2020-01-15T00:00:00",
   "end-of-semester":"2020-05-12T00:00:00",
   "cards":[
      {
         "type":"recent-transactions",
         "header-title":"Transactions",
         "header-body":"Your recent dining dollar transactions",
         "data":[
            {
               "location":"Canteen Vending API - Charge",
               "date":"2020-02-17T23:39:00",
               "balance":0,
               "amount":-1.45
            },
            {
               "location":"Canteen Vending API - Charge",
               "date":"2020-02-17T01:52:00",
               "balance":1.45,
               "amount":-2.99
            },
            {
               "location":"Canteen Vending API - Charge",
               "date":"2020-02-17T01:49:00",
               "balance":4.44,
               "amount":-1.49
            },
            {
               "location":"Canteen Vending API - Charge",
               "date":"2020-02-11T16:32:00",
               "balance":5.93,
               "amount":-14.07
            },
            {
               "location":"Canteen Vending API - Charge",
               "date":"2020-02-10T22:01:00",
               "balance":20,
               "amount":-4.06
            }
         ]
      }
   ]
}
```


dining/views.py
```diff
@@ -0,0 +1,86 @@
...
+		self.pennid = "76627463" # request.user.username
+		self.uid = Account.objects.filter(pennid=self.pennid)[0].id
+		json = self.balance()
+
+		start, end = self.get_semester_start_end()
+		json["start-of-semester"] = start
+		json["end-of-semester"] = end
+
+		json["cards"] = []
+		json["cards"].append(self.recent_transactions_card())
```
davish:
```suggestion
		json["cards"] = [self.recent_transactions_card()]
```
and delete the line above to do it in one line

astrike30: Thank so much, will do all of that. There are also a few things that Dom has requested be added quickly, so I'll add those in to and commit again.

studentlife/utils.py
```diff
@@ -0,0 +1,63 @@
...
+    return events
+
+
+def date_iso_eastern(date):
+
+    date = date.replace(tzinfo=None)
+    eastern = pytz.timezone("US/Eastern")
+    date = eastern.localize(date)
+
+    print(date)
```
davish: remember to take out all print statements before merging 😄

dining/views.py
```diff
@@ -0,0 +1,90 @@
...
+# Create your views here.
+
+
+class Dashboard(APIView):
+    def get(self, request, format=None):
+        if not self.request.user.is_authenticated:
+            return HttpResponseForbidden()
+
+        self.pennid = "76627463"  # request.user.username
+        self.uid = Account.objects.filter(pennid=self.pennid)[0].id
```
davish:
```suggestion
        try:
             self.uid = Account.objects.get(pennid=self.pennid).id
        catch Account.DoesNotExist:
            pass # some code here in case the pennID doesn't exist
```

davish: also realized the indentation on that suggestion may be messed up :/ but generally if you're looking for one element it's better to use `get()` than `filter()[0]`.

dining/views.py
```diff
@@ -0,0 +1,90 @@
...
+        start, end = self.get_semester_start_end()
+        json["start-of-semester"] = date_iso_eastern(start)
+        json["end-of-semester"] = date_iso_eastern(end)
+
+        json["cards"] = {"recent-transactions": self.recent_transactions_card()}
+
+        return JsonResponse(json)
+
+    def balance(self):
+        balance = DiningBalance.objects.filter(account=self.uid).order_by("-created_at")
```
davish: Like I mentioned above, I think it would be better to pass `uid` into this function instead of setting it on `self` and then referencing the instance.

studentlife/utils.py
astrike30: 🤦🏻‍♂️

PR #15, opened 2020-03-01
Group Booking Logic (Libcal)

r3khaan:
Added route under GroupViewSet.py for booking a room with your group at /groups/<group:id>/book-room/

Currently, you need to pass the following params: "rooms", "start", "end", "lid".

Here is an example post request: {
        "room": "16993",
	"start": "2020-03-06T18:00:00-0500",
	"end": "2020-03-06T18:30:00-0500",
	"lid": "2587"
}


gsr_booking/booking_logic.py
```diff
@@ -0,0 +1,255 @@
...
+                break
+
+        if is_wharton:  # huntsman reservation
+            return {
+                "complete_success": False,
+                "partial_success": False,
+                "error": "Unable to book huntsman rooms yet",
+            }
+        else:  # lib reservation
+            # Find the first timeslot to book for (next_start, next_end)
```
davish: eventually will be nice to pull out `book_wharton` and `book_libcal` into separate functions for readability.

gsr_booking/booking_logic.py
```diff
@@ -0,0 +1,255 @@
...
+            START_DATE = datetime.datetime.strptime(start, self.DATE_FORMAT_STR)
+            END_DATE = datetime.datetime.strptime(end, self.DATE_FORMAT_STR)
+            next_start = START_DATE
+            next_end = min(END_DATE, next_start + datetime.timedelta(hours=self.MAX_SLOT_HRS))
+
+            # loop through each member, and attempt to book on their behalf
+            bookings = {}
```
davish:
```suggestion
            bookings = dict()
```
just for clarity -- `{}` can also be a set in python.

gsr_booking/booking_logic.py
```diff
@@ -0,0 +1,255 @@
...
+            failed_members = []  # store the members w/ failed bookings in here
+
+            for member in members:
+                if next_end - next_start < datetime.timedelta(hours=0.1):
```
davish: would be nice to have this timedelta as a constant instead of hardcoded in, for easy access/change.

gsr_booking/booking_logic.py
```diff
@@ -0,0 +1,255 @@
...
+                    break
+                if member["user__email"] is "":
```
davish: document that this represents a missing email and so this user can't be used for libcal.

gsr_booking/booking_logic.py
```diff
@@ -0,0 +1,255 @@
+import requests
+import datetime
+import math
+import random
+
+
+class BookingLogic:
+
+    DATE_FORMAT_STR = "%Y-%m-%dT%H:%M:%S%z"
```
davish: I think this is just the ISO datetime format? if so i think you can use this function: https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat

gsr_booking/booking_logic.py
```diff
@@ -0,0 +1,255 @@
...
+        # makes a request to labs api server to book rooms, and returns result json if successful
+        members = group.get_pennkey_active_members()
+        if len(members) < 1:
+            return {
+                "complete_success": False,
+                "partial_success": False,
+                "error": "No members in the group have enabled their pennkey to be used for group bookings",
+            }
+
+        # randomize order, and put requester first in the order
```
davish: Don't know if you want this to be a TODO to work on a more nuanced booking algorithm

gsr_booking/booking_logic.py
```diff
@@ -0,0 +1,255 @@
...
+                    )
+                elif self.is_fatal_error(error):
+                    result_json = self.construct_bookings_json_obj(
+                        bookings, lid, room, END_DATE, next_start, next_end, error
+                    )
+                    return result_json
+                else:
+                    failed_members.append(member)
+
+            # if unbooked slots still remain and not all booking requests succeeded, loop through each member again
```
davish: I know we talked about this before, but looking at the code, just wondering why we don't get how much credit a user has left before attempting a booking in the first place? Seems like this might just be duplicating some work. Feel like the algorithm would be shorter if we didn't do the "blind" booking first.

gsr_booking/booking_logic.py
```diff
@@ -0,0 +1,255 @@
...
+                    lid, member["user__email"]
+                )
+                remaining_credit_hours = self.MAX_SLOT_HRS - used_credit_hours
+                rounded_remaining_credit_hours = math.floor(2 * remaining_credit_hours) / 2
+                if success and remaining_credit_hours >= self.MIN_SLOT_HRS:
+                    next_end = min(
+                        END_DATE,
+                        next_start + datetime.timedelta(hours=rounded_remaining_credit_hours),
+                    )
+                    (success, error) = self.book_room_for_user(
```
davish: A more pythonic thing to do would probably be to throw an exception and then using a try/except block instead of returning the error in a tuple

gsr_booking/booking_logic.py
```diff
@@ -0,0 +1,255 @@
...
+                        room,
+                        lid,
+                        next_start.strftime(self.DATE_FORMAT_STR),
+                        next_end.strftime(self.DATE_FORMAT_STR),
+                        member["user__email"],
+                    )
+                    if success:
```
davish: I think if we keep the blind booking above, this logic here should be factored out into a function to avoid duplication.

gsr_booking/booking_logic.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def is_fatal_error(self, error):
+        # consider all errors fatal unless its a daily limit error
+        if error is not None:
+            if "daily limit".lower() not in error.lower():
+                return False
+            return True
+        return False
+
+    def construct_bookings_json_obj(
```
davish: Would be cool to write a docstring here that shows what a booking json object looks like, just so devs don't have to construct it in their head.

gsr_booking/booking_logic.py
```diff
@@ -0,0 +1,255 @@
...
+                    reservation_hours = (to_date - from_date).total_seconds() / 3600
+                    if str(reservation["lid"]) == lid and reservation_hours > 0.1:
+                        used_credit_hours += reservation_hours
+                return (True, used_credit_hours)
+        except requests.exceptions.RequestException as e:
+            print(e)
+        return (False, 0)
+
+    def split_booking(self, start, end, pennkey, booked):
+        # splits a booking into smaller bookings (of min_slot_hrs) for displaying to user
```
davish:
```suggestion
        """
        splits a booking into smaller bookings (of min_slot_hrs) for displaying to user
        """
```
Docstrings in python are triple-quoted strings :)

gsr_booking/booking_logic.py
```diff
@@ -0,0 +1,255 @@
...
+                "booked": booked,
+            }
+            if pennkey is not None:
+                booking_obj["pennkey"] = pennkey
+            bookings.append(booking_obj)
+            temp_start = temp_end
+            temp_end += datetime.timedelta(hours=self.MIN_SLOT_HRS)
+        return bookings
+
+    def book_room_for_user(self, room, lid, start, end, email):
```
davish: making a class that's a subclass of `Exception` and throwing that would be better than returning the error in a tuple.

gsr_booking/booking_logic.py
```diff
@@ -0,0 +1,255 @@
...
+        # tries to make a booking for an individual user, and returns success or not (and the error) in a tuple
+        if lid is "1":
+            return False  # does not support huntsman booking yet
```
davish:
```suggestion
            return False, None  # does not support huntsman booking yet
```
So it works with the tuple and you don't get weird restructuring errors

gsr_booking/booking_logic.py
r3khaan: By document, do you mean comment it? Or do you mean return it to the user?

gsr_booking/booking_logic.py
r3khaan: I know we have talked about a more nuanced booking algorithm before, but I'm curious what that entails specifically?

gsr_booking/booking_logic.py
r3khaan: Agreed - although I have a feeling they are going to share a decent amount of code

gsr_booking/booking_logic.py
r3khaan: The algorithm would be shorter, but slower on average if we skipped the "blind" booking. This is because getting reservations for a user takes ~2x times longer than making a booking for that user. Also, although including "blind" booking does increase the worse case time for booking, in the average case we would only need to get reservation credits for a few users(assuming that some people in the group have not booked anything).

I talked with Josh and he suggested that in the future we can use the booking credit data stored in SQL. Although it only captures the bookings made through pennmobile, we would be able to get reservation credits almost instantly, allowing us to remove the blind booking part and drastically decease time to book for group.

gsr_booking/booking_logic.py
r3khaan: Got it, so would I then have add a try/catch in the function that calls the `book_room_for_user` function? (so that I can capture the error and return it in JSON format)

r3khaan: Nevermind, just saw your other comment I understand

gsr_booking/booking_logic.py
r3khaan: The problem is I need to update the local variables (`next_start, next_end, bookings`), so if I made a function it would have maybe 1 line (for finding the key). Any other ideas?

gsr_booking/booking_logic.py
davish: I mean comment -- sorry for the ambiguity there 🙃

gsr_booking/booking_logic.py
davish: good point! I think i had a question somewhere in here about what the purpose of a blind booking was? versus just always checking how much availability someone has.

gsr_booking/models.py
```diff
@@ -59,13 +59,27 @@ class Group(models.Model):
...
+    MEMBER = "M"
+
     def __str__(self):
         return "{}: {}".format(self.pk, self.name)
 
     def has_member(self, user):
         memberships = GroupMembership.objects.filter(group=self, accepted=True)
         return memberships.all().filter(user=user).exists()
 
+    def is_admin(self, user):
```
davish: maybe calling this `has_admin` to match `has_member` above? at first it looked like this was a method on GroupMembership, not a Group.

gsr_booking/views.py
```diff
@@ -3,15 +3,30 @@
...
+from gsr_booking.csrfExemptSessionAuthentication import CsrfExemptSessionAuthentication
+from gsr_booking.models import Group, GroupMembership, UserSearchIndex
+from gsr_booking.serializers import (
+    GroupMembershipSerializer,
+    GroupSerializer,
+    UserSerializer,
+    GroupBookingRequestSerializer,
+)
+
+from .booking_logic import (
```
davish: relative imports can get yucky sometimes, so best to stick to `from gsr_booking.booking_logic ...`

gsr_booking/booking_logic.py
r3khaan: For sure - I did respond to that question in another comment but GitHub thinks its "outdated" for some reason, so I'll post my answer here:

The algorithm would be shorter, but slower on average if we skipped the "blind" booking. This is because getting reservations for a user takes ~2x times longer than making a booking for that user. Also, although including "blind" booking does increase the worse case time for booking, in the average case we would only need to get reservation credits for a few users (assuming that some people in the group have not booked anything).

I talked with Josh and he suggested that in the future we can use the booking credit data stored in SQL. Although it only captures the bookings made through pennmobile, we would be able to get reservation credits almost instantly, allowing us to remove the blind booking part and drastically decease time to book for group.

PR #17, opened 2020-03-23
Added support for simultaneously booking multiple rooms

r3khaan:
Sample request JSON looks like this:
`{
   "room_bookings":[
      {
         "room": 7282,
         "start":"2020-03-25T21:30:00.000Z",
         "end":"2020-03-25T22:30:00.000Z",
         "lid": 1090
      },
      {
         "room": 7282,
         "start":"2020-03-25T22:30:00.000Z",
         "end":"2020-03-25T23:30:00.000Z",
         "lid": 1090
      }
   ]
}`

### DIFF
--- a/gsr_booking/serializers.py
+++ b/gsr_booking/serializers.py
@@ -5,21 +5,21 @@ from rest_framework import serializers
 
 User = get_user_model()
 
-
-class GroupBookingRequestSerializer(serializers.Serializer):
+class GroupRoomBookingRequestSerializer(serializers.Serializer):
     lid = serializers.IntegerField()
     room = serializers.IntegerField()
 
     start = serializers.DateTimeField()
     end = serializers.DateTimeField()
 
-    group_id = serializers.IntegerField()
-
     is_wharton = serializers.SerializerMethodField()
 
     def get_is_wharton(self, obj):
         return obj["lid"] == 1
 
+class GroupBookingRequestSerializer(serializers.Serializer):
+    room_bookings = serializers.ListField(child=GroupRoomBookingRequestSerializer())
+    group_id = serializers.IntegerField()
 
 class MiniUserSerializer(serializers.ModelSerializer):
     class Meta:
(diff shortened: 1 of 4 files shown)


gsr_booking/serializers.py
```diff
@@ -6,21 +6,23 @@
...
-    group_id = serializers.IntegerField()
-
     is_wharton = serializers.SerializerMethodField()
 
     def get_is_wharton(self, obj):
         return obj["lid"] == 1
 
 
+class GroupBookingRequestSerializer(serializers.Serializer):
+    room_bookings = serializers.ListField(child=GroupRoomBookingRequestSerializer())
```
davish:
```suggestion
    room_bookings = GroupRoomBookingRequestSerializer(many=True)
```
I think this will work in the same way, and it's a bit less verbose. It also matches the ModelSerializer format a bit more for if/when we decide to save all of this to the DB.

gsr_booking/views.py
```diff
@@ -267,27 +267,22 @@ def invites(self, request, pk):
...
         if not booking_serialized.is_valid():
             return Response(status=400)
 
         booking_data = booking_serialized.data
 
-        group = get_object_or_404(Group, pk=booking_data["group_id"])
+        group = get_object_or_404(Group, pk=pk)
+
+        # must be admin of the group to book
         if not group.has_member(request.user) or not group.has_admin(request.user):
```
davish: A way to do this in one database call instead of two would be to query for the GroupMembership for this group and user, and then check that that membership `is_admin`.

gsr_booking/views.py
```diff
@@ -267,27 +267,22 @@ def invites(self, request, pk):
             ).data
         )
 
-    @action(detail=True, methods=["post"], url_path="book-room")
-    def book_room(self, request, pk):
-        request.data.update({"group_id": pk})
+    @action(detail=True, methods=["post"], url_path="book-rooms")
+    def book_rooms(self, request, pk):
```
davish: Remember to add a docstring here so that documentation shows up on ReDoc!

PR #20, opened 2020-12-31
Accounts

ArmaanT:
This PR will configure all the account information needed for student-life.
[ch2093]


studentlife/urls.py
```diff
@@ -10,7 +10,7 @@
 urlpatterns = [
     path("", include("gsr_booking.urls")),
     path("admin/", admin.site.urls),
-    path("accounts/", include("accounts.urls", namespace="accounts")),
+    path("user/", include("user.urls")),
```
ArmaanT, line 13: We'll still need the accounts endpoints since those come from DLA

user/serializers.py
```diff
@@ -0,0 +1,32 @@
+from django.contrib.auth import get_user_model
+from rest_framework import serializers
+
+from .models import NotificationToken, Profile
```
ArmaanT, line 4: Try to use absolute imports instead of relative ones, they make things a little bit clearer to read. We actually use an addon to flake8 that enforces this, but I guess it wasn't set up on this repo

gsr_booking/migrations/0002_auto_20210129_1527.py
```diff
@@ -0,0 +1,38 @@
+# Generated by Django 3.1.4 on 2021-01-29 20:27
```
ArmaanT, line 1: How did this end up being generated? Looks like the code was updated but someone didn't run makemigrations on it?

ArmaanT (commented):
Can't request changes because this is my own PR, but just a couple of nits @judtinzhang

gsr_booking/migrations/0002_auto_20210129_1527.py
judtinzhang: yea im not entirely sure. I ran makemigrations to populate the database for the user app, but i guess it also made a new migration for the gsr_booking app as well ?

PR #22, opened 2021-02-04
Server/Laundry backend migration

judtinzhang:
Moved all laundry routes
Updated laundry_data CSV to incorporate total washers and dryers
Created laundry api_wrapper file based off of penn-sdk-python repo
Added additional fields to models in order to store data rather than
having to access the Penn API every time


laundry/tests.py
```diff
@@ -0,0 +1,4 @@
+# from django.test import TestCase
```
ArmaanT, line 1: This file can be deleted (and tests can go in `tests/laundry/xyz.py` :wink: )

setup.cfg
```diff
@@ -5,7 +5,7 @@ inline-quotes = double
 
 [isort]
 default_section = THIRDPARTY
-known_first_party = studentlife
+known_first_party = studentlife, dining, gsr_booking, user
```
ArmaanT, line 8: don't forget to add laundry to this list

laundry/urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path
+
+from laundry import views
```
ArmaanT, line 3: doing `from laundry.views import (Halls, etc)` would be a bit cleane

laundry/views.py
```diff
@@ -0,0 +1,124 @@
...
+from laundry.api_wrapper import Laundry
+from laundry.models import LaundrySnapshot
+from laundry.serializers import LaundrySnapshotSerializer
+
+
+laundry = Laundry()
+
+
+# should i just use the regular API view with a GET header?
+class Halls(generics.ListAPIView):
```
ArmaanT, line 19: I think for all of these you should be fine using just APIView and a `get` method. I think you could also use JsonResponse here instead of Response

laundry/views.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class HallInfo(generics.ListAPIView):
+    def list(self, request, hall_id):
+        try:
+            return Response(laundry.hall_status(int(hall_id)))
+        except ValueError:
+            return Response({"error": "Invalid hall id passed to server."})
+        except HTTPError:
+            return Response({"error": "The laundry api is currently unavailable."})
```
ArmaanT, line 34: These should return a non-200 status code. Probably a 404 for the ValueError and maybe a 503 for the HTTPError

laundry/views.py
```diff
@@ -0,0 +1,124 @@
...
+            )
+            snapshots = snapshots.union(new_snapshots)
+
+        return snapshots
+
+    def list(self, request, hall_id):
+
+        snapshots = self.get_queryset(hall_id)
+
+        print(snapshots.count())
```
ArmaanT, line 70: remove print

laundry/data/laundry_data.csv
```diff
@@ -0,0 +1,53 @@
+0,Bishop White,Quad,5faec7e9-a4aa-47c2-a514-950c03fac460
```
ArmaanT, line 1: I think that it could make sense to have a LaundryMachine model that contains all of these fields. That way the machines are stored in our DB and user's favorited machines can be a direct link to a LaundryMachine. Do you guys think that makes sense? We should definitely talk more about this

laundry/api_wrapper.py
```diff
@@ -0,0 +1,200 @@
+import csv
+import os
+
+import pkg_resources
+import requests
+from bs4 import BeautifulSoup
+
+
+LAUNDRY_URL = os.environ.get("LAUNDRY_URL", "http://suds.kite.upenn.edu")
+ALL_URL = "{}/?location=".format(LAUNDRY_URL)
```
ArmaanT, line 10: I know this was directly copied from LAS but it can be optimized to `= f"{LAUNDRY_URL}/?location="`

laundry/api_wrapper.py
```diff
@@ -0,0 +1,200 @@
...
+    def __init__(self):
+        self.busy_dict = {
+            "LowBusyNightColor": "Low",
+            "LowBusyDayColor": "Low",
+            "MediumLowBusyColor": "Medium",
+            "MediumHighBusyColor": "High",
+            "HighBusyColor": "Very High",
+            "NoDataBusyColor": "No Data",
+        }
+        self.days = ["Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday", "Sunday"]
```
ArmaanT, line 25: @kevc528 I guess you're not the only one to have Monday first here

laundry/api_wrapper.py
```diff
@@ -0,0 +1,200 @@
...
+from bs4 import BeautifulSoup
+
+
+LAUNDRY_URL = os.environ.get("LAUNDRY_URL", "http://suds.kite.upenn.edu")
+ALL_URL = "{}/?location=".format(LAUNDRY_URL)
+
+# USAGE_BASE_URL from LAS does not work
+
+
+class Laundry(object):
```
ArmaanT, line 15: Depending on our discussion about a LaundryMachine model (see other comment) this may need to be rewritten slightly

ArmaanT (changes requested):

laundry/views.py
judtinzhang: is there a particular difference between jsonresponse and response other than response being a DRF thing?

ArmaanT: Actually, I take that back. JsonResponse is a [native django class](https://docs.djangoproject.com/en/3.1/ref/request-response/#jsonresponse-objects) not a DRF one, and it doesn't have the ability to negotiate a renderer. Response is the way to go!

laundry/management/commands/uuid_migration.py
```diff
@@ -0,0 +1,20 @@
+import csv
```
ArmaanT, line 1: I would rename this to something like `load_laundry_rooms` or something similar

laundry/models.py
```diff
@@ -0,0 +1,25 @@
+import uuid
+
+from django.db import models
+
+
+class LaundrySnapshot(models.Model):
+    hall_id = models.IntegerField()
```
ArmaanT, line 7: This could be a foreign key to `LaundryRoom`

laundry/models.py
```diff
@@ -0,0 +1,25 @@
...
+    date = models.DateTimeField(auto_now_add=True)
+    available_washers = models.IntegerField()
+    available_dryers = models.IntegerField()
+    total_washers = models.IntegerField()
+    total_dryers = models.IntegerField()
```
ArmaanT, line 12: Since these don't change over time, we could probably just stick them in the `LaundryRoom` model

laundry/models.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def __str__(self):
+        return f"Hall No. {self.hall_id} | {self.date.date()}"
+
+
+class LaundryRoom(models.Model):
```
ArmaanT, line 18: Could you add a comment here describing why we have a uuid field? Just so we remember what it's for at a glance

laundry/views.py
```diff
@@ -0,0 +1,160 @@
...
+            }
+        )
+
+
+class Preferences(APIView):
+    def get(self, request):
+
+        # tests if user is logged-in
+        if not request.user.is_authenticated:
+            return Response({"rooms": []})
```
ArmaanT, line 131: You should be able to use the DRF `IsAuthenticated` permission to simplify this

laundry/views.py
```diff
@@ -0,0 +1,160 @@
...
+
+        preferences = Profile.objects.get(user=request.user).preferences.all()
+
+        hall_ids = [x.hall_id for x in preferences]
```
ArmaanT, line 135: You can combine this into a single `values_list` (I think that's the function) from a queryset

laundry/views.py
```diff
@@ -0,0 +1,160 @@
...
+
+        # returns all hall_ids in a person's preferences
+        return Response({"rooms": hall_ids})
+
+    def post(self, request):
+
+        # tests if user is logged-in
+        if not request.user.is_authenticated:
+            return Response({"rooms": []})
```
ArmaanT, line 144: Ditto on the IsAuthenticated permission

laundry/views.py
```diff
@@ -0,0 +1,160 @@
...
+
+        profile = Profile.objects.get(user=request.user)
```
ArmaanT, line 146: You should also be able to just do `request.user.profile` or something like that to immediately get the user's profile

user/migrations/0001_initial.py
```diff
@@ -1,4 +1,4 @@
-# Generated by Django 3.1.5 on 2021-02-03 22:59
+# Generated by Django 3.1.5 on 2021-01-29 21:36
```
ArmaanT, line 1: Remember not to modify existing migrations

user/migrations/0002_auto_20210205_1851.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 3.1.5 on 2021-02-05 23:51
```
ArmaanT, line 1: This and 0003 should be combined

laundry/api_wrapper.py
```diff
@@ -0,0 +1,229 @@
...
+        self.hall_to_link = {}
+        self.id_to_hall = {}
+        self.id_to_location = {}
+        self.hall_id_list = []
+        self.create_hall_to_link_mapping()
+
+    def create_hall_to_link_mapping(self):
+        try:
+            # makes migration when there are no rooms
+            if LaundryRoom.objects.count() == 0:
```
ArmaanT, line 38: I don't think we should have this code. Since the command only needs to be run once, then the db will always have the info

laundry/api_wrapper.py
ArmaanT: Okay, I think now that we have all of the information about laundry rooms stored in the DB, I think we refactor this so that we don't create a new class, but instead just provide functions that reference the DB instead of maintaining a duplicate set of information in memory

ArmaanT (changes requested):

laundry/views.py
```diff
@@ -0,0 +1,157 @@
...
+        except HTTPError:
+            return Response({"error": "The laundry api is currently unavailable."}, status=503)
+
+
+class HallInfo(APIView):
+    def get(self, request, hall_id):
+        try:
+            return Response(hall_status(int(hall_id)))
+        except ValueError:
+            return Response({"error": "Invalid hall id passed to server."}, status=404)
```
ArmaanT, line 29: This could be simplified into a LaundryRoom [get_or_404](https://docs.djangoproject.com/en/3.1/topics/http/shortcuts/#get-object-or-404) and then `hall_status` could just take in a LaundryRoom object. That way we move away from just passing ints around and hoping a room exists with that id

laundry/views.py
```diff
@@ -0,0 +1,157 @@
...
+
+        # start is beginning of day, end is 27 hours after start
+        start = make_aware(datetime.datetime(year=now.year, month=now.month, day=now.day))
+        end = start + datetime.timedelta(hours=27)
+
+        # filters for LaundrySnapshots within timeframe
+        try:
+            room = LaundryRoom.objects.get(hall_id=hall_id)
+        except LaundryRoom.DoesNotExist:
+            raise ValueError("No hall with id %s exists." % hall_id)
```
ArmaanT, line 52: You could use a get_or_404 here too

laundry/views.py
```diff
@@ -0,0 +1,157 @@
...
+            return Response(hall_status(int(hall_id)))
+        except ValueError:
+            return Response({"error": "Invalid hall id passed to server."}, status=404)
+        except HTTPError:
+            return Response({"error": "The laundry api is currently unavailable."}, status=503)
+
+
+class HallUsage(APIView):
+
+    serializer_class = LaundrySnapshotSerializer
```
ArmaanT, line 36: I don't think you're using a `LaundrySnapshotSerializer` here, right?

ArmaanT (changes requested):

laundry/views.py
```diff
@@ -0,0 +1,167 @@
...
+    def post(self, request):
+
+        profile = request.user.profile
+
+        # clears all previous preferences in many-to-many
+        profile.laundry_preferences.clear()
+
+        hall_ids = request.data["rooms"]
+
+        for hall_id in hall_ids:
```
ArmaanT, line 160: This loop should probably be in a try/catch that returns a 400 if something goes wrong. Normally DRF would handle that for us, but since we're not using serializers to do this update we have to do it manually. Down the line, we should probably revisit this to make it more DRF-y

ArmaanT (approved):

PR #26, opened 2021-03-24
server/misc. penn-related data on penn mobile migration + dining fixes

judtinzhang:
/news/ - gets front page article from the DP
/calendar/ - gets newest university events within a 2 week range
/events/<type>/ - gets list of all events by type
/homepage - [WIP] returns cells for homepage features, still missing GSR bookings + portal + new features announcement

quick fix to dining request -- catches Timeout error


penndata/models.py
```diff
@@ -0,0 +1,13 @@
+from django.db import models
+
+
+class Event(models.Model):
+    event_type = models.CharField(max_length=255)
```
ArmaanT, line 5: Are there any limits as to what type an event may be?

penndata/views.py
```diff
@@ -0,0 +1,203 @@
...
+    """
+    list:
+    Return a list of events belonging to the type (events/?type=)
+    """
+
+    permission_classes = [AllowAny]
+    serializer_class = EventSerializer
+
+    def get_queryset(self):
+        return Event.objects.filter(event_type=self.kwargs["type"])
```
ArmaanT, line 133: I think this will 500 when `type` isn't provided as a query arg. If that's the case, I think it may be better to do something like `self.kwargs.get("type", "")` so it has a default value

ArmaanT (changes requested):
Two minor comments

ArmaanT (approved):

PR #33, opened 2021-06-28
Gsr migration

judtinzhang:


gsr_booking/api_wrapper.py
```diff
@@ -0,0 +1,331 @@
...
+from django.utils import timezone
+
+
+BASE_URL = "https://libcal.library.upenn.edu"
+API_URL = "https://api2.libcal.com"
+WHARTON_URL = "https://apps.wharton.upenn.edu/gsr/api/v1/"
+
+# unbookable rooms
+LOCATION_BLACKLIST = set([3620, 2636, 2611, 3217, 2637, 2634])
+ROOM_BLACKLIST = set([7176, 16970, 16998, 17625])
```
ArmaanT, line 15: These can be `= {3620, 2636, ...}`

gsr_booking/api_wrapper.py
```diff
@@ -0,0 +1,331 @@
...
+
+class APIError(ValueError):
+    pass
+
+
+class WhartonLibWrapper:
+    def request(self, *args, **kwargs):
+        """Make a signed request to the libcal API."""
+
+        headers = {"Authorization": "Token {}".format(settings.WHARTON_TOKEN)}
```
ArmaanT, line 26: Should use an f-string here

gsr_booking/api_wrapper.py
```diff
@@ -0,0 +1,331 @@
...
+
+        self.expiration = timezone.localtime() + datetime.timedelta(seconds=response["expires_in"])
+        self.token = response["access_token"]
+
+    def request(self, *args, **kwargs):
+        """Make a signed request to the libcal API."""
+        if not self.token:
+            self.get_token()
+
+        headers = {"Authorization": "Bearer {}".format(self.token)}
```
ArmaanT, line 118: same f-string comment here and elsewhere

gsr_booking/views.py
```diff
@@ -323,3 +336,121 @@ def book_rooms(self, request, pk):
...
+                context["gid"] = room["cid"]
+                context["rooms"] = [
+                    {"room_name": x["name"], "id": x["id"], "availability": x["availability"]}
+                    for x in room["rooms"]
+                ]
+                response.append(context)
+            return Response(response)
+
+
+class BookRoom(APIView):
```
ArmaanT, line 396: All of the views from here on should probably have an IsAuthenticated permission check, right?

penndata/views.py
```diff
@@ -149,13 +156,21 @@ def __init__(self, myType, myInfo=None, myWeight=0):
         def getCell(self):
             return {"type": self.type, "info": self.info}
 
+    def get_gsr_invites(self, username):
```
ArmaanT, line 159: You should just pass the user to this method instead of the username which requires a check to get the underlying user

ArmaanT (changes requested):
A few minor comments. Looks good though

PR #34, opened 2021-06-30
made post-save hook for creating profiles

judtinzhang:


tests/penndata/test_views.py
```diff
@@ -109,13 +107,13 @@ def test_first_response(self):
         self.assertEqual(res_json[3]["type"], "laundry")
```
ArmaanT, line 107: I know it's not part of the PR, but a few lines above this you can use `self.assertIn` to simplify some of the tests

ArmaanT (approved):
One minor comment that's not really related to the PR. Code changes LGTM!

PR #37, opened 2021-09-03
Polls Backend

judtinzhang:
First iteration of Penn Mobile Polls Backend. Completed everything except getting user school years


portal/serializers.py
```diff
@@ -0,0 +1,89 @@
...
+        instance = Poll(**validated_data, user=self.context["request"].user)
+        # ensuring user cannot create an admin comment upon creation
+        instance.admin_comment = ""
+        instance.save()
+        return instance
+
+    def update(self, instance, validated_data):
+        user = self.context["request"].user
+        # checks if user is allowed to update or not
+        if user != instance.user:
```
ArmaanT, line 22: This should be done in a DRF permission

portal/serializers.py
```diff
@@ -0,0 +1,89 @@
...
+            "id",
+            "poll",
+            "choice",
+        )
+
+    def update(self, instance, validated_data):
+        user = self.context["request"].user
+        # checks to see if user can update the option
+        if user != instance.poll.user:
+            raise serializers.ValidationError(detail={"detail": "User not allowed to update."})
```
ArmaanT, line 46: ditto on DRF permission comment

portal/serializers.py
```diff
@@ -0,0 +1,89 @@
...
+        return super().update(instance, validated_data)
+
+
+class RetrievePollSerializer(serializers.ModelSerializer):
+
+    options = PollOptionSerializer(source="polloption_set", many=True)
```
ArmaanT, line 52: does this need a `read_only=True` or something similar? I thought django tends to complain if you don't do that

portal/serializers.py
```diff
@@ -0,0 +1,89 @@
...
+        fields = ("id", "poll_option")
+
+    def create(self, validated_data):
+        # adds poll and user to the vote
+        instance = PollVote(
+            **validated_data,
+            user=self.context["request"].user,
+            poll=validated_data["poll_option"].poll
+        )
+        instance.save()
```
ArmaanT, line 71: Same comment on modifying validated data instead of making the pollvote object yourself

portal/serializers.py
```diff
@@ -0,0 +1,89 @@
...
+        return instance
+
+    def update(self, instance, validated_data):
+        user = self.context["request"].user
+        # makes sure only user can adjust vote (admins cannot change votes)
+        if user != instance.user:
+            raise serializers.ValidationError(detail={"detail": "User not allowed to update."})
```
ArmaanT, line 78: DRF permission comment here too

portal/serializers.py
```diff
@@ -0,0 +1,89 @@
...
+        return super().update(instance, validated_data)
+
+
+class RetrievePollVoteSerializer(serializers.ModelSerializer):
```
ArmaanT, line 82: This seems like it could probably be merged with the previous serializer, right?

portal/urls.py
```diff
@@ -0,0 +1,28 @@
...
+    UpdatePollOption,
+    UpdatePollVote,
+)
+
+
+router = routers.DefaultRouter()
+
+router.register(r"polls/view", RetrievePolls)
+urlpatterns = [
+    path("", include(router.urls)),
```
ArmaanT, line 20: It doesn't really matter, but I tend to prefer splitting the router and regular URLs like we do in [OHQ](https://github.com/pennlabs/office-hours-queue/blob/master/backend/ohq/urls.py)

portal/views.py
```diff
@@ -0,0 +1,174 @@
...
+
+    permission_classes = [IsAuthenticated]
+    queryset = Poll.objects.all()
+    serializer_class = RetrievePollSerializer
+
+    @action(detail=False, methods=["get"])
+    def browse(self, request):
+        # filters for all possible available polls, then filters for all polls answered
+        # by user, and finally returns the difference in poll sets
+        polls_available = Poll.objects.filter(expire_date__gte=timezone.localtime(), approved=True)
```
ArmaanT, line 31: Should this be a timezone.localtime() or timezone.now()?

portal/views.py
```diff
@@ -0,0 +1,174 @@
...
+        ).values_list("poll", flat=True)
+        return Response(
+            self.serializer_class(polls_available.exclude(id__in=polls_answered), many=True).data
+        )
+
+    @action(detail=False, methods=["get"])
+    def review(self, request):
+        # checks if user is admin
+        if not request.user.is_superuser:
+            raise PermissionDenied()
```
ArmaanT, line 43: This should be in a DRF permission as well

portal/views.py
```diff
@@ -0,0 +1,174 @@
...
+    permission_classes = [IsAuthenticated]
+    queryset = Poll.objects.all()
+    serializer_class = RetrievePollSerializer
+
+    @action(detail=False, methods=["get"])
+    def browse(self, request):
+        # filters for all possible available polls, then filters for all polls answered
+        # by user, and finally returns the difference in poll sets
+        polls_available = Poll.objects.filter(expire_date__gte=timezone.localtime(), approved=True)
+        polls_answered = PollVote.objects.filter(
```
ArmaanT, line 32: I feel like there should be a better way of doing this by I don't know SQL/the ORM well enough to say for sure

portal/views.py
```diff
@@ -0,0 +1,174 @@
...
+            self.serializer_class(polls_available.exclude(id__in=polls_answered), many=True).data
+        )
+
+    @action(detail=False, methods=["get"])
+    def review(self, request):
+        # checks if user is admin
+        if not request.user.is_superuser:
+            raise PermissionDenied()
+        # returns list of polls that still need to be approved
+        return Response(self.serializer_class(Poll.objects.filter(approved=False), many=True).data)
```
ArmaanT, line 45: I think you can also do this by making a custom get_queryset function that returns different results depending on the view

portal/views.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+class RetrievePollVotes(APIView):
+    """Retrieve history of polls and their statistics"""
+
+    def get(self, request):
+        if not request.user.is_authenticated:
+            raise PermissionDenied()
```
ArmaanT, line 53: ditto on DRF permission comment

portal/views.py
```diff
@@ -0,0 +1,174 @@
...
+            return "SEAS"
+        elif "sas" in email:
+            return "SAS"
+        elif "nursing" in email:
+            return "Nursing"
+        else:
+            return "Other"
+
+
+class CreatePoll(generics.CreateAPIView):
```
ArmaanT, line 115: I'm pretty sure most of these Create/Update/List/ individual views can be merged into viewsets to keep things a bit more concise

ArmaanT (changes requested):

portal/serializers.py
judtinzhang: it did not throw an error for me after testing haha. should i put it just in case?

portal/serializers.py
judtinzhang: instead of permissions, i just changed all of them via the get_queryset() function. basically, the queryset only contains the objects that they created

portal/views.py
judtinzhang: i've generally always used localtime, otherwise in testing the times are all UTC. im not sure if this is the same as production but just wanted to be safe

portal/serializers.py
judtinzhang: i think they should be kept separate. the RetrievePollVoteSerializer doesn't take in an "<int: id>" as the id of a Poll or PollOption, and basically requires me to create the object on the spot, whereas the PollVoteSerializer doesn't. And i need the RetrievePollVoteSerializer for displaying user history

portal/permissions.py
```diff
@@ -0,0 +1,43 @@
...
+class TimeSeriesPermission(permissions.BasePermission):
+    """Permission that checks for Time Series access (only creator of Poll and admins)"""
+
+    def has_permission(self, request, view):
+        poll = Poll.objects.filter(id=view.kwargs["id"])
+        # checks if poll exists
+        if poll.exists():
+            # only poll creator and admin can access
+            return (
+                poll.first().user == request.user and request.user.is_authenticated
```
ArmaanT, line 41: I'm pretty sure you don't need the `request.user.is_authenticated` part, right? Because poll users have to be an actual user

portal/serializers.py
```diff
@@ -0,0 +1,131 @@
...
+        fields = ("id", "poll_options", "created_date")
+        read_only_fields = ("created_date",)
+
+    def create(self, validated_data):
+
+        options = validated_data["poll_options"]
+        poll = options[0].poll
+        # checks if user has already voted
+        if PollVote.objects.filter(user=self.context["request"].user, poll=poll).exists():
+            print(PollVote.objects.filter(user=self.context["request"].user))
```
ArmaanT, line 95: remove print statement

portal/serializers.py
```diff
@@ -0,0 +1,131 @@
...
+        # check if poll options are all from same Poll
+        for option in options:
+            if option.poll != poll:
+                raise serializers.ValidationError(
+                    detail={"detail": "Voting options are from different Polls"}
+                )
+        # checks if user belongs to target population
+        if not any(
+            i in list(poll.target_populations.all().values_list("id", flat=True))
+            for i in get_user_populations(self.context["request"].user)
```
ArmaanT, line 113: I'm pretty sure this is an inefficient way of doing this check. Looks like you iterate over all poll target populations for each user population, right?

portal/views.py
```diff
@@ -0,0 +1,192 @@
...
+    partial_update:
+    Update certain fields in the Poll.
+    Only user who creates Poll can edit it
+
+    destroy:
+    Delete a Poll.
+    """
+
+    permission_classes = [OwnerPermission]
+    queryset = Poll.objects.all()
```
ArmaanT, line 52: I'm not sure, but I think you don't need to/shouldn't define this if you also use `get_queryset`

portal/views.py
```diff
@@ -0,0 +1,192 @@
...
+    Update certain fields in the membership.
+    Only specify the fields that you want to change.
+
+    destroy:
+    Delete a Poll Option.
+    """
+
+    permission_classes = [OwnerPermission]
+    serializer_class = PollOptionSerializer
+    queryset = PollOption.objects.all()
```
ArmaanT, line 144: same queryset comment

portal/views.py
```diff
@@ -0,0 +1,192 @@
...
+
+class PollVoteTimeSeries(APIView):
+    """Returns time series of all votes for a particular poll"""
+
+    permission_classes = [TimeSeriesPermission]
+
+    def get(self, request, id):
+        return Response(
+            {
+                "time_series": PollVote.objects.filter(poll__id=id)
```
ArmaanT, line 186: Does this not need to be wrapped in a serializer or something?

ArmaanT (changes requested):

portal/serializers.py
ArmaanT: I'm not entirely sure what you mean by taking in an id for the Poll(Option. I was thinking that you just move the create method from the previous serializer to this one and it should more or less work. We should probably talk more about this in person on Friday

portal/serializers.py
ArmaanT: Bump on this. Would something like `poll.target_populations.filter(id__in=get_user_populations(self.context["request"].user).count == 0` not work?

ArmaanT (changes requested):
Couple of comments

PR #44, opened 2021-09-27
Create populate.py

annypan:


judtinzhang (changes requested):
Great work! A few edits though. Sorry for being a stickler here, but just want to make sure that the database entires represent ALL cases because this file will be used extensively by iOS, android, and portal developers. 

Just a few general notes on top of the comments:
1. there should be ~1, 2 polls without poll options. in this case, they would have an `approved=False`. for the other polls, they should always have poll options
2. there should be ~1, 2 polls without target populations, also should be not approved. for the others, they should always have target populations. 
3. if you could add some diversity for school and graduation year in the votes and poll creation, that would be amazing just to make sure all our cases are covered for analytics purposes
4. if you have questions feel free to reach out! hopefully we can get this up by friday because thats when the other developers will start working on it :)

judtinzhang (approved):
LGTM!

PR #45, opened 2021-10-04
Added recent poll and limit poll options to 5

Kmanbear:
Added recent poll and limit poll options to 5, added tests

### DIFF
--- a/portal/views.py
+++ b/portal/views.py
@@ -95,12 +95,20 @@ class Polls(viewsets.ModelViewSet):
         )
 
 
-class RetrievePollVotes(APIView):
+class RetrievePollVotes(viewsets.ModelViewSet):
     """Retrieve history of polls and their statistics"""
 
     permission_classes = [IsAuthenticated | IsSuperUser]
+    serializer_class = RetrievePollSerializer
 
-    def get(self, request):
+    def get_queryset(self):
+        return (
+            Poll.objects.all()
+            if self.request.user.is_superuser
+            else Poll.objects.filter(user=self.request.user)
+        )
+
+    def list(self, request):
 
         # filters for all polls that user voted in
         serializer = RetrievePollVoteSerializer(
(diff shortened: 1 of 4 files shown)


portal/views.py
```diff
@@ -130,10 +138,14 @@ def get(self, request):
...
             history_list.append(context)
-        for entry in history_list:
-            entry["poll_statistics"] = get_demographic_breakdown(entry["poll"]["id"])
         return Response(sorted(history_list, key=lambda i: i["poll"]["expire_date"], reverse=True))
 
+    @action(detail=False, methods=["get"])
+    def recent(self, request, pk=None):
+        recent_poll_vote = PollVote.objects.filter(user=request.user).latest("created_date")
+        recent_poll = recent_poll_vote.poll
+        return Response(RetrievePollSerializer(recent_poll).data)
```
judtinzhang, line 147: I think this needs to be a `RetrievePollVoteSerializer`. On the `list` function it is a bit funky but I'm using that `RetrievePollVoteSerializer`, and then calling `RetrievePollSerializer`, but then converting it into the form of `RetrievePollVoteSerializer` for consistency

tests/portal/test_polls.py
```diff
@@ -188,12 +226,26 @@ def test_update_vote(self):
...
-        self.assertEqual({}, res_json[1]["poll_statistics"][0]["breakdown"])
+
+    def test_recent_poll(self):
+        # answer poll
+        payload_1 = {"poll_options": [self.p1_op1_id]}
+        self.client.post("/portal/votes/", payload_1)
+        response = self.client.get(reverse("portal:poll-history-recent"))
+        res_json = json.loads(response.content)
+        # assert that payload1 poll answered is most recent
+        self.assertEquals(self.p1_op1_id, res_json["id"])
```
judtinzhang, line 244: i am not entirely sure this test case works (check out the github actions). i think it would be better to get the `id` from filtering for the most recent `PollVote` object, and then comparing it to `res_json['id']`

tests/portal/test_polls.py
```diff
@@ -188,12 +226,26 @@ def test_update_vote(self):
...
+        # answer another poll
+        payload_2 = {"poll_options": [self.p4_op1_id]}
+        self.client.post("/portal/votes/", payload_2)
+        response2 = self.client.get(reverse("portal:poll-history-recent"))
+        res_json2 = json.loads(response2.content)
+        # assert newly answered poll is most recent
+        self.assertEquals(self.p4_op1_id, res_json2["id"])
```
judtinzhang, line 251: ditto on last comment about test cases

judtinzhang (changes requested):
Good work! Just a few minor edits 👍

PR #55, opened 2021-10-29
Single person group

lantingchiang:


lantingchiang:
Somehow calling get_or_create() in a migration file appears to have not called the save() function in Group (which should have added a GroupMembership instance with the user as the admin of the group), so I manually create a GroupMembership instance if we create a Group instance

judtinzhang (approved):
lgtm!

PR #57, opened 2021-10-31
Portal post: create tests for posts

annypan:


backend/portal/views.py
```diff
@@ -213,3 +215,45 @@ def get(self, request, id):
...
+        return Post.objects.all()
+
+    @action(detail=False, methods=["get"])
+    def browse(self, request):
+        """
+        Returns a list of all posts that are targeted at the current user
+        Admins sees all the posts(?)
+        """
+        posts = (
+            Post.objects.all()
```
judtinzhang, line 251: we might need to change this query, as the get_user_populations function is not necessarily working atm

backend/portal/views.py
```diff
@@ -213,3 +215,45 @@ def get(self, request, id):
...
+            if request.user.is_superuser
+            else Post.objects.filter(
+                Q(target_populations__in=get_user_populations(request.user)),
+                start_date__lte=timezone.localtime(),
+                expire_date__gte=timezone.localtime(),
+            )
+        )
+        return Response(PostSerializer(posts.distinct().order_by("start_date"), many=True).data)
```
judtinzhang, line 259: same here, might also need to order_by approved

judtinzhang (changes requested):

PR #63, opened 2021-11-12
Portal posts: create status route

annypan:


backend/portal/views.py
```diff
@@ -277,3 +277,37 @@ def review(self, request):
...
             ).data
         )
+
+    @action(detail=False, methods=["get"])
+    def status(self, request):
+        # awaiting approval: no admin comment, not approved
+        posts_awaiting_approval = PostSerializer(
+            Post.objects.filter(
+                (Q(admin_comment=None) | Q(admin_comment="")),
+                Q(user=request.user),
```
judtinzhang, line 287: for the `user=request.user`, i don't think you need to wrap the Q object around it. we only use Q objects for advanced searches (or's, __in's, etc.)

backend/portal/views.py
```diff
@@ -277,3 +277,37 @@ def review(self, request):
...
+                approved=False,
+                user=request.user,
+            ),
+            many=True,
+        ).data
+
+        # revision: have admin comment, not approved
+        posts_revision = PostSerializer(
+            Post.objects.filter(
+                ~Q(admin_comment=None), ~Q(admin_comment=""), Q(user=request.user), approved=False
```
judtinzhang, line 297: ditto on the `user=request.user` stuff

judtinzhang (changes requested):
just one tiny thing, nice work and lmk if it runs on ur local server!

PR #66, opened 2022-01-01
Portal Backend <> Clubs & Platform Integration

judtinzhang:
- Completely anonymous voting (hash of device id + poll id)
- Integrated with Clubs --> clubs create Polls and users have access to clubs they're admins of
- Integrated with Platform --> Population targeting Polls and poll statistics


backend/portal/logic.py
```diff
@@ -6,26 +9,60 @@
...
+        ]
     )
-    # year = user.profile.expected_graduation.year if user.profile.expected_graduation else None
-    year = None
-    # checks if year is in the target population
-    year_id = (
-        TargetPopulation.objects.get(population=year).id
-        if TargetPopulation.objects.filter(population=year).exists()
-        else -1
+    content.extend(
```
annypan, line 57: Just curious: if the user satisfies more than one criterion (eg: targeted school & targeted major), will the same `TargetPopulation` be added twice to `content`?

annypan: Oh nvm I think I got it: according to the current design, each TargetPopulation object can only specify one and only one kind. Could you confirm that my understanding is correct? @judtinzhang

PR #83, opened 2022-02-26
Notifications V1

tuneerroy:


backend/gsr_booking/management/commands/send_gsr_reminders.py
```diff
@@ -0,0 +1,46 @@
...
+from user.notifications import send_push_notif
+
+
+class Command(BaseCommand):
+    help = "Sends reminders for the GSR Bookings."
+
+    def handle(self, *args, **kwargs):
+        # iterate through all reservations scheduled for the next 30 minutes
+        for reservation in Reservation.objects.filter(
+            start__gt=timezone.now(), start__lte=timezone.now() + datetime.timedelta(minutes=30)
```
judtinzhang, line 17: for this query, if you could change to within 10 minutes instead of 30, as we will be notifying uses on the 20th and 50th minute every hour. also, you can also filter for `is_cancelled=False` instead of doing the if statement on line 19

backend/gsr_booking/management/commands/send_gsr_reminders.py
```diff
@@ -0,0 +1,46 @@
...
+            token = (
+                NotificationToken.objects.filter(user=creator, kind=NotificationToken.KIND_IOS)
+                .exclude(token="")
+                .first()
+            )
+
+            if token:
+                # skip user if their setting = disabled for GSR Notifs
+                setting = token.objects.filter(service=SERVICE_GSR_BOOKING).first()
+                if setting and not setting.enabled:
```
judtinzhang, line 34: for this if statement, i'm not sure if it also takes care of if the setting is non-existent. maybe the conditional could be like `if not setting or not settings.enabled`

backend/user/management/commands/send_shadow_notifs.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class Command(BaseCommand):
+    help = "Sends shadow notifs to all or targeted users."
+
+    def handle(self, *args, **kwargs):
+        send_to_all = input("Send to all (yes/no)? ").lower() == "yes"
+        # get list of targeted users if not to everyone
+        if not send_to_all:
+            usernames = input("Target users (comma-delimitted): ").split(",")
```
judtinzhang, line 16: for lines 16 and 18, do you mind also adding an example of the input we are looking for? just want to add this cause i don't want others to continuously ask us what the inputs for this looks like haha

backend/user/models.py
```diff
@@ -15,12 +15,46 @@ class NotificationToken(models.Model):
...
+        (SERVICE_PENN_COURSE_PLAN, "Penn Course Plan"),
+        (SERVICE_PENN_COURSE_REVIEW, "Penn Course Review"),
+        (SERVICE_PENN_MOBILE, "Penn Mobile"),
+        (SERVICE_GSR_BOOKING, "GSR Booking"),
+        (SERVICE_DINING, "Dining"),
+        (SERVICE_UNIVERSITY, "University"),
+        (SERVICE_LAUNDRY, "Laundry"),
+    )
+
+    # user = models.ForeignKey(User, on_delete=models.CASCADE)
```
judtinzhang, line 52: maybe delete this line haha

backend/user/serializers.py
```diff
@@ -1,19 +1,53 @@
...
+            return setting
+        else:
+            return super().create(validated_data)
+
 
 class ProfileSerializer(serializers.ModelSerializer):
     class Meta:
         model = Profile
-        fields = ("expected_graduation", "degrees", "laundry_preferences", "dining_preferences")
+        # removed "expected_graduation" and "degrees"
```
judtinzhang, line 49: i think its fine to remove this comment as well

backend/user/views.py
```diff
@@ -21,3 +31,95 @@ class UserView(generics.RetrieveUpdateAPIView):
...
+        return NotificationSetting.objects.filter(token__user=self.request.user)
+
+    @action(detail=True, methods=["get"])
+    def check(self, request, pk=None):
+        """
+        Returns whether the user wants notification for specified service.
+        :param pk: service name
+        """
+        # style
+        pk = pk.upper().replace("-", "_")
```
judtinzhang, line 74: i think that this probably won't be necessary. reason being is that the iOS team won't know about this niche thing. although it does look nicer, funcitonality-wise i think it might be best to just leave it as PENN_MOBILE in the url

judtinzhang (changes requested):
overall looks good! just some minor changes. and after that we can work on tests together and then get this show on the road :D

judtinzhang (approved):
LGTM 👍 Just gotta make some test cases then we are good to go!

backend/user/notifications.py
```diff
@@ -0,0 +1,56 @@
...
+    payload = Payload(content_available=True, custom=body)
+    notifications = [Notification(token.token, payload) for token in tokens]
+    topic = "org.pennlabs.PennMobile" + (".dev" if isDev else "")
+    client.send_notification_batch(notifications=notifications, topic=topic)
+
+
+def get_client(isDev):
+    """Creates and returns APNsClient based on iOS credentials"""
+
+    auth_key_path = os.path.join(
```
judtinzhang, line 47: Can you make a Shortcut story on this line? Don't know if this will be a problem, but not sure if the file path is the same in production as it was in labs-api-server. If notifications works fine, that's great! Otherwise, this might be the root of the bug

judtinzhang (approved):
LGTM!

PR #100, opened 2022-04-11
Fix Notification Settings

tuneerroy:


backend/user/management/commands/fix_user_settings.py
```diff
@@ -0,0 +1,28 @@
...
+        NotificationSetting.objects.all().delete()
+        self.stdout.write("All settings deleted!")
+
+        # create the proper settings for all users
+        SERVICE_OPTIONS = NotificationSetting.SERVICE_OPTIONS
+
+        for user in User.objects.all():
+            token, _ = NotificationToken.objects.get_or_create(user=user)
+            for service, _ in SERVICE_OPTIONS:
+                setting = NotificationSetting.objects.filter(token=token, service=service).first()
```
judtinzhang, line 24: i think lines 24,25,26 can be changed into a `update_or_create` function to save an extra query in the DB.
https://docs.djangoproject.com/en/4.0/ref/models/querysets/#update-or-create

judtinzhang (changes requested):

judtinzhang (approved):

PR #105, opened 2022-04-18
Celery + Redis

tuneerroy:
- [x] set up celery
- [x] add delayed notifications
- [x] set up redis caching 
- [x] refactor notifications.py file
- [x] redo painful testing

[Shortcut](https://app.shortcut.com/pennlabs/story/3633/notifications-v2-django-celery)


backend/pennmobile/settings/production.py
```diff
@@ -20,7 +20,7 @@
 
 # Sentry settings
 SENTRY_URL = os.environ.get("SENTRY_URL", "")
-sentry_sdk.init(dsn=SENTRY_URL, integrations=[DjangoIntegration()])
+sentry_sdk.init(dsn=SENTRY_URL, integrations=[CeleryIntegration(), DjangoIntegration()])
 
 # DLA settings
 PLATFORM_ACCOUNTS = {"ADMIN_PERMISSION": "penn_mobile_admin"}
```
mureytasroc, line 26: Since you're using Redis in prod now, you might as well use it for [Django's cache backend](https://docs.djangoproject.com/en/4.0/topics/cache/):

```
CACHES = {
    "default": {
        "BACKEND": "django_redis.cache.RedisCache",
        "LOCATION": REDIS_URL,
        "OPTIONS": {
            "CLIENT_CLASS": "django_redis.client.DefaultClient",
        },
    }
}
```

([source](https://github.com/pennlabs/penn-courses/blob/master/backend/PennCourses/settings/production.py#L32) in PCx settings).

Then if you actually use caching in Django, you should add the following to settings.development as well:

```
CACHES = {
    "default": {
        "BACKEND": "django.core.cache.backends.dummy.DummyCache",
    }
}
```

([source](https://github.com/pennlabs/penn-courses/blob/master/backend/PennCourses/settings/development.py#L16) in PCx settings). Note that we use dummy caching in dev even though we have Redis in our dev environment (for Celery), to enforce that we don't actually rely on Django's caching for important storage (i.e. the Django cache can be purged at any time, causing possible slowdowns, but without causing any change in semantics).

For an example of how we purge Django cache entries while leaving the Celery part of the cache alone, see [here](https://github.com/pennlabs/penn-courses/blob/master/backend/review/management/commands/clearcache.py).

mureytasroc (commented):
Added comments about some Celery/Redis stuff. Everything else looks good to me, although I didn't review changes related to notification tokens or send_shadow_notifs.py (don't have context on those).

backend/user/views.py
```diff
@@ -91,13 +96,18 @@ class NotificationAlertView(APIView):
...
-        isDev = request.data["isDev"] if "isDev" in request.data else False
+        usernames = request.data.get("users", [self.request.user.username])
+        service = request.data.get("service", None)
+        title = request.data.get("title", None)
+        body = request.data.get("body", None)
+        delay = request.data.get("delay", None)
+
+        if None in [service, title, body]:
+            return Response(
+                {"detail": "Missing required parameters."}, status=status.HTTP_400_BAD_REQUEST
```
judtinzhang, line 107: ditto on the status codes

backend/user/views.py
```diff
@@ -109,14 +119,32 @@ def post(self, request):
...
         )
 
+        # unpack list of tuples
+        if tokens:
+            success_users, tokens, devs = zip(*tokens)
+        else:
+            success_users, devs = list(), list()
+        failed_users = list(set(usernames) - set(success_users))
+
         if len(tokens) == 1:
```
judtinzhang, line 132: hmm, I know we used this everywhere, but it is getting a little confusing with all of the `tokens` variables around. I think this one refers to the `tokens` variable after the zip, and the one on L126 refers to the QuerySet. Do you think we should change or nah?

backend/user/notifications.py
```diff
@@ -4,29 +4,41 @@
...
 from apns2.payload import Payload
+from celery import shared_task
 
 
 # taken from the apns2 method for batch notifications
 Notification = collections.namedtuple("Notification", ["token", "payload"])
 
 
+# celery delayed tasks
+def send_delayed_push_notif(token, title, body, countdown, isDev=False):
```
judtinzhang, line 15: nitpick here, I think in the `views.py` file you referenced `countdown` and `delay`. I would suggest using one or the other

backend/user/views.py
```diff
@@ -109,14 +119,32 @@ def post(self, request):
...
+                send_delayed_push_notif(tokens[0], title, body, delay, devs[0])
+            else:
+                send_push_notif(tokens[0], title, body, devs[0])
         elif len(tokens) > 1:
-            send_push_notif_batch(tokens, title, body, isDev)
+            # ensure that tokens are the same type
+            if any(dev != devs[0] for dev in devs):
+                return Response(
+                    {"detail": "Includes both non-dev and dev users."},
+                    status=status.HTTP_400_BAD_REQUEST,
```
judtinzhang, line 142: ditto on status

backend/user/management/commands/send_shadow_notifs.py
```diff
@@ -54,13 +54,16 @@ def handle(self, *args, **kwargs):
                     user__username__in=usernames,
                 )
                 .exclude(token="")
+                .values_list("user__username", "token")
```
judtinzhang, line 57: do we need to update this to include `dev` mode as well?

backend/user/notifications.py
```diff
@@ -4,29 +4,41 @@
...
     client = get_client(isDev)
     alert = {"title": title, "body": body}
     payload = Payload(alert=alert, sound="default", badge=0)
     topic = "org.pennlabs.PennMobile" + (".dev" if isDev else "")
-    client.send_notification(token.token, payload, topic)
+    client.send_notification(token, payload, topic)
 
 
+@shared_task(name="tasks.send_push_notif_batch")
 def send_push_notif_batch(tokens, title, body, isDev=False):
```
judtinzhang, line 35: this code and `send_shadow_push_notif_batch` looks about exactly the same except for one line. might be a good idea to just have 1 method and have a parameter for `send_shadow` or something, and it defaults to `False`

backend/user/views.py
```diff
@@ -109,14 +119,32 @@ def post(self, request):
...
         if len(tokens) == 1:
-            send_push_notif(tokens[0], title, body, isDev)
+            if delay:
+                send_delayed_push_notif(tokens[0], title, body, delay, devs[0])
+            else:
+                send_push_notif(tokens[0], title, body, devs[0])
         elif len(tokens) > 1:
-            send_push_notif_batch(tokens, title, body, isDev)
+            # ensure that tokens are the same type
+            if any(dev != devs[0] for dev in devs):
```
judtinzhang, line 139: I think this line might be a bit slow (10000+ tokens). Also, this line will fail for any batch notification that includes Jongmin, because he has 2 notification tokens (one dev and one not) and not sure if your query differentiates between that. I would suggest removing this line and making the query on L114 filter for a `dev=False`, unless otherwise specified.

backend/user/views.py
```diff
@@ -109,14 +119,32 @@ def post(self, request):
                 notificationsetting__enabled=True,
             )
             .exclude(token="")
+            .values_list("user__username", "token", "dev")
         )
 
+        # unpack list of tuples
```
judtinzhang, line 125: I think these few lines and the query on L114 are also duplicated in `send_shadow_notifs.py`. Might be a good idea to de-duplicate them by putting it in `notifications.py`

judtinzhang (changes requested):
Very cool :D Just a couple changes / possible design choices. Feel free to pick and choose

backend/pennmobile/settings/base.py
```diff
@@ -156,6 +156,13 @@
     ],
 }
 
+# Redis for Celery & Caching
+REDIS_URL = os.environ.get("REDIS_URL", "redis://localhost:6379/1")
+
+# Celery
+CELERY_BROKER_URL = REDIS_URL
```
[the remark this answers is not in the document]
tuneerroy: used for Celery config (auto-read)

backend/user/management/commands/clearcache.py
```diff
@@ -0,0 +1,38 @@
+import logging
```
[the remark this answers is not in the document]
tuneerroy: I'm not too sure of any cases where we'll need to clear cache (first time using caching tbh), but since we are caching, I think it's fine to keep the file, just in case we need to but we don't want to mess with celery. Charley pointed the file out to me when recommending we set up caching.

backend/pennmobile/settings/base.py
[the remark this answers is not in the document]
judtinzhang: brief comment here on showing that it is actually used in celery.py

backend/user/management/commands/clearcache.py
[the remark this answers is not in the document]
judtinzhang: Ask charley when this is used

.gitignore
```diff
@@ -20,5 +20,8 @@ htmlcov/
 # Test database
 db.sqlite3
 
+# Redis
```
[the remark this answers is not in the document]
tuneerroy: whenever Redis is stopped on the local env, it creates the file

backend/Pipfile
```diff
@@ -39,6 +39,10 @@ pillow = "*"
 boto3 = "*"
 apns2 = "*"
 numpy = "*"
+click = "==8.0.4"
```
judtinzhang, line 42: Very cool very cool

backend/user/notifications.py
```diff
@@ -4,41 +4,83 @@
...
-    alert = {"title": title, "body": body}
-    payload = Payload(alert=alert, sound="default", badge=0)
-    topic = "org.pennlabs.PennMobile" + (".dev" if isDev else "")
-    client.send_notification(token.token, payload, topic)
+    :param usernames: list of usernames to send notifications to or 'None' if to all
+    :param service: service to send notifications for or 'None' if ignoring settings
+    :param title: title of notification
+    :param body: body of notification
+    :param delay: delay in seconds before sending notification
+    :param isShadow: whether to send a shadow notification
```
judtinzhang, line 25: camel case 🦨

judtinzhang (approved):
LGTM :D

backend/user/notifications.py
tuneerroy: whatHaveIDone!

backend/pennmobile/settings/production.py
```diff
@@ -20,7 +21,16 @@
...
+
```
tuneerroy, line 28: (for DevOps to look at)

k8s/main.ts
```diff
@@ -12,6 +12,17 @@ export class MyChart extends PennLabsChart {
     const backendImage = "pennlabs/penn-mobile-backend"
     const frontendImage = "pennlabs/penn-mobile-frontend"
 
+    new RedisApplication(this, 'redis', { deployment: { tag: '4.0' } });
```
tuneerroy, line 15: (for DevOps to look at)

backend/pennmobile/celery.py
```diff
@@ -0,0 +1,25 @@
+import os
```
tuneerroy, line 1: (for DevOps to look at)

joyliu-q (approved):
DevOps related changes in `k8s` and the `settings/production.py` looks good to me!

PR #118, opened 2022-09-25
Notif Bug Fixes

tuneerroy:


backend/user/notifications.py
```diff
@@ -44,10 +44,11 @@ def send_shadow_push_notif_batch(tokens, body, isDev=False):
 def get_client(isDev):
     """Creates and returns APNsClient based on iOS credentials"""
 
-    auth_key_path = os.path.join(
-        os.path.dirname(os.path.dirname(os.path.abspath(__file__))), "ios_key.p8"
+    auth_key_path = os.environ.get(
+        "IOS_KEY_PATH",  # for dev purposes
```
judtinzhang, line 48: Even if we are using client for non-dev purposes, will this still work?

backend/user/serializers.py
```diff
@@ -11,14 +11,14 @@ class Meta:
...
-        token = NotificationToken.objects.filter(user=validated_data["user"]).first()
+        token_obj = NotificationToken.objects.filter(user=validated_data["user"]).first()
 
-        if token:
-            # if token already exists, just update it
-            token.kind = validated_data["kind"]
-            token.token = validated_data["token"]
-            token.save()
-            return token
+        if token_obj:
```
judtinzhang, line 16: Hm, now that I am looking at this, I think I messed up last semester lol. The code here is basically just redefining the `update(...)` function inherited from the serializer. What I am thinking of now:

1. Removing this code, and overriding the `update(self, instance, validated_data)` function to just error if they try to update the user, and the call the `super().update(...)` function. This update function will handle all PATCH requests. Example of overriding [here](https://github.com/pennlabs/penn-mobile/blob/master/backend/portal/serializers.py#L107)
2. Erroring out in the `create(...)` function whenever there already exists a token for the user

judtinzhang (changes requested):

backend/user/notifications.py
tuneerroy: If the env variable is not set, then it will default to the path laid out, so it should work just like before.

judtinzhang (approved):

PR #140, opened 2022-10-23
Fix Notifications Dev Flag

tuneerroy:


backend/Pipfile
```diff
@@ -40,7 +40,7 @@ boto3 = "*"
 apns2 = "*"
 numpy = "*"
 click = "==8.0.4"
-celery = "5"
```
judtinzhang, line 43: Looks good! Per discussion, this is because there was no reason to have celery be hard-set to 5.0. 

Also, let's remove this from the other PR since we are making the change here.

backend/user/views.py
```diff
@@ -95,14 +95,15 @@ def post(self, request):
         title = request.data.get("title", None)
         body = request.data.get("body", None)
         delay = request.data.get("delay", 0)
+        is_dev = request.data.get("isDev", False)
```
judtinzhang, line 98: nit: I don't know if this is already set in stone for iOS, because if so we can leave it. But maybe we should snake case `isDev` if there's not a huge hassle.

backend/user/views.py
```diff
@@ -95,14 +95,15 @@ def post(self, request):
...
 
         if None in [service, title, body]:
             return Response({"detail": "Missing required parameters."}, status=400)
         if service not in dict(NotificationSetting.SERVICE_OPTIONS):
             return Response({"detail": "Invalid service."}, status=400)
 
         success_users, failed_users = send_push_notifications(
-            usernames, service, title, body, delay
+            usernames, service, title, body, delay=delay, is_dev=is_dev
```
judtinzhang, line 106: Is there a reason why we defined the variables `delay=delay` and `is_dev=is_dev` here? Not sure if needed.

judtinzhang (commented):
Slight nits, but overall looks good!

backend/user/management/commands/send_shadow_notifs.py
```diff
@@ -28,24 +28,27 @@ def add_arguments(self, parser):
         parser.add_argument("message", type=str, help="JSON-formatted message to send")
 
         # optional argument
-        parser.add_argument("--usernames", type=str, help="list of usernames")
+        parser.add_argument("--users", type=str, help="list of usernames")
```
judtinzhang, line 31: nit: Suggest writing list of pennkeys just to be explicit to non-django developers

judtinzhang (approved):
LGTM!

PR #143, opened 2022-10-28
Refactor Get Venues

tuneerroy:
Instead of requesting data from UPenn every time the `/api/dining/venues` endpoint is, we gather the data from the DiningMenu objects we already hold.

### DIFF
--- a/backend/dining/api_wrapper.py
+++ b/backend/dining/api_wrapper.py
@@ -58,8 +58,8 @@ class DiningAPIWrapper:
 
     def get_venues(self):
         venues = Venue.objects.all()
-        res = []
-        for venue in venues:
+        res = [None] * len(venues)
+        for i, venue in enumerate(venues):
             venues_obj = {"id": venue.venue_id, "name": venue.name, "image": venue.image_url}
 
             days = []
@@ -78,7 +78,7 @@ class DiningAPIWrapper:
                 ]
                 days.append(days_obj)
             venues_obj["days"] = days
-            res.append(venues_obj)
+            res[i] = venues_obj
         return res
 
     def load_menu(self, date=timezone.now().date()):
--- a/backend/dining/api_wrapper.py
+++ b/backend/dining/api_wrapper.py
@@ -57,57 +57,29 @@ class DiningAPIWrapper:
             raise APIError("Dining: Connection timeout")
 
     def get_venues(self):
-        results = []
-        venues_route = OPEN_DATA_ENDPOINTS["VENUES"]
-        response = self.request("GET", venues_route)
-        if response.status_code != 200:
-            raise APIError("Dining: Error connecting to API")
-        venues = response.json()["result_data"]["campuses"]["203"]["cafes"]
-        for key, value in venues.items():
-            # Cleaning up json response
-            venue = Venue.objects.filter(venue_id=key).first()
-            value["name"] = venue.name
-            value["image"] = venue.image_url if venue else None
+        venues = Venue.objects.all()
+        res = []
+        for venue in venues:
+            venues_obj = {"id": venue.venue_id, "name": venue.name, "image": venue.image_url}
 
-            value["id"] = int(key)
-            remove_items = [
-                "cor_icons",
-                "city",
-                "state",
-                "zip",
-                "latitude",
-                "longitude",
-                "description",
-                "message",
-                "eod",
-                "timezone",
-                "menu_type",
-                "menu_html",
-                "location_detail",
-                "weekly_schedule",
-            ]
-            [value.pop(item) for item in remove_items]
-            for day in value["days"]:
-                day.pop("message")
-                removed_dayparts = set()
-                for i in range(len(day["dayparts"])):
-                    daypart = day["dayparts"][i]
-                    [daypart.pop(item) for item in ["id", "hide"]]
-                    if not daypart["starttime"]:
-                        removed_dayparts.add(i)
-                        continue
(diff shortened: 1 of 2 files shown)


backend/dining/api_wrapper.py
```diff
@@ -57,57 +57,29 @@ def request(self, *args, **kwargs):
...
+            for day_offset in range(7):
+                days_obj = dict()
+
+                date = datetime.datetime.today() + datetime.timedelta(days=day_offset)
+                date = date.strftime("%Y-%m-%d")
+                days_obj["date"] = date
+
+                # get starttimes and endtimes from DiningMenu objects
+                days_obj["dayparts"] = [
+                    {"starttime": menu.start_time, "endtime": menu.end_time, "label": menu.service}
```
tuneerroy, line 76: Do we want to change the object keys to match our variables' names and have iOS change it on their side?

judtinzhang: For now, let's keep the variable names as is, because changes to iOS and Android actually take a while to propagate. Later we should make this more Django native but now it'll be fine haha

backend/dining/api_wrapper.py
```diff
@@ -57,57 +57,29 @@ def request(self, *args, **kwargs):
...
+        for i, venue in enumerate(venues):
+            venues_obj = {"id": venue.venue_id, "name": venue.name, "image": venue.image_url}
+
+            days = []
+            # get info for a week
+            for day_offset in range(7):
+                days_obj = dict()
+
+                date = timezone.now() + datetime.timedelta(days=day_offset)
+                date = date.strftime("%Y-%m-%d")
```
judtinzhang, line 71: Hm, because we are re-using the `date` variable here, might be confusing. Suggest combining lines 71 and 72 into `days_obj["date"] = date.strftime...`

backend/dining/api_wrapper.py
```diff
@@ -155,17 +127,17 @@ def load_menu(self, date=timezone.now().date()):
                 dining_menu.save()
 
     def load_stations(self, station_response):
-        # Store stations into list
-        stations = list()
-        for station_data in station_response:
+        stations = [None] * len(station_response)
+        for i, station_data in enumerate(station_response):
```
judtinzhang, line 131: Cool :D

backend/tests/dining/test_views.py
```diff
@@ -35,6 +37,52 @@ def json(self):
         return Mock(json.load(data), 200)
 
 
+def mock_request_raise_error(*args, **kwargs):
```
judtinzhang, line 40: Wowowow error testing haha

backend/tests/dining/test_views.py
```diff
@@ -35,6 +37,52 @@ def json(self):
...
+
+def mock_request_post_error(*args, **kwargs):
+    class Mock:
+        def json(self):
+            return {"error": None}
+
+    return Mock()
+
+
+class TestTokenAndRequest(TestCase):
```
judtinzhang, line 52: nit: you can override the `setUp` function in `TestCase` so that you don't have to re-define `DiningAPIWrapper()` for every function call. This `setUp` function will be re-invoked for every `test_*` function you have so each test will be independent of each other.

backend/tests/dining/test_views.py
```diff
@@ -45,16 +93,13 @@ def test_get(self):
         response = self.client.get(reverse("venues"))
         for entry in response.json():
             self.assertIn("name", entry)
-            self.assertIn("address", entry)
             self.assertIn("days", entry)
             self.assertIn("image", entry)
             self.assertIn("id", entry)
             for day in entry["days"]:
                 self.assertIn("date", day)
-                self.assertIn("status", day)
```
judtinzhang, line 54: Hm, can you double check with iOS and Android (in the f22 channel) to make sure they don't use this `status` variable? Preferably this PR will be backwards compatible. I think this `status` just shows "open/closed" but that can easily be derived by them by just checking the start and end times. (same with `address`)

backend/tests/dining/test_views.py
```diff
@@ -93,6 +138,16 @@ def test_get_date(self):
         response = self.client.get("/dining/menus/2022-10-04/")
         self.try_structure(response.json())
 
+    @mock.patch("requests.request", mock_request_raise_error)
+    def test_skip_vanue(self):
```
judtinzhang, line 142: nit: vanue lol

backend/tests/dining/test_views.py
```diff
@@ -93,6 +138,16 @@ def test_get_date(self):
...
+        Venue.objects.all().delete()
+        Venue.objects.create(venue_id=747, name="Skip", image_url="URL")
+        wrapper = DiningAPIWrapper()
```
judtinzhang, line 145: Hm, I'm wondering if there is a way to assert an Error in Python Testcase (similar to 121 Java testing)?

judtinzhang (changes requested):
Looking good! Minor changes

backend/tests/dining/test_views.py
tuneerroy: https://stackoverflow.com/questions/4319825/python-unittest-opposite-of-assertraises
:(

backend/tests/dining/test_views.py
tuneerroy: ok. confirmed and we should be good.

judtinzhang (approved):
LGTM!

PR #146, opened 2022-11-06
Remove Dev from NotificationToken Model

tuneerroy:


judtinzhang (approved):
Just to confirm -- whenever a dev wants to test, they will update their token in the DB, and then will send this `is_dev` flag to the alerts view?

PR #160, opened 2023-02-05
Unique Portal Visits

vcai122:
* Adds `eventcount/` endpoint under portal app to return the number of unique interactions for either polls or posts
* Adds test cases for new endpoint


tuneerroy (approved):

judtinzhang:
Good stuff guys! Also before pushing, would suggest renaming PR to something more formal + leave a description!

PR #167, opened 2023-02-14
GSR Commands

tuneerroy:
- [x] Add/remove/reset people to/from groups
- [x] Retrieve the number of GSR reservations ever made in a given time frame (can be for specific group)
- [x] Determine total time of all GSR reservations ever made
- [x] Retrieve data on all Group GSRs
- [x] Unique users booking GSRs
- [ ] Write test cases

Shortcuts:
https://app.shortcut.com/pennlabs/story/3791/gsr-update-v1-fix-internal-group-gsr-tool
https://app.shortcut.com/pennlabs/story/3919/scripts-to-write-for-business-development

### DIFF
--- a/backend/gsr_booking/management/commands/change_group.py
+++ b/backend/gsr_booking/management/commands/change_group.py
@@ -10,25 +10,34 @@ User = get_user_model()
 
 class Command(BaseCommand):
     help = """
-    Adds users to a group.
-    --reset flag
-    Sets all group members to be the users specified in the command.
+    Adds/remove users to a group.
+
+    --remove flag for users to be removed from the group
+    --reset flag to set group to be all users specified in command
+
+    Note: --remove and --reset are mutually exclusive
     """
 
     def add_arguments(self, parser):
         parser.add_argument("usernames", type=str, help="list of pennkeys")
         parser.add_argument("group", type=str, help="group name")
 
-        # optional flag to set group to be all users
+        # optional flags
+        parser.add_argument("--remove", type=bool, default=False)
         parser.add_argument("--reset", type=bool, default=False)
 
     def handle(self, *args, **kwargs):
         usernames = kwargs["usernames"].split(",")
         group = kwargs["group"]
+        remove = kwargs["remove"]
         reset = kwargs["reset"]
 
+        if not usernames:
+            self.stdout.write("Error: no users specified")
+            return
+
         if not (group := Group.objects.filter(name=group).first()):
-            self.stdout.write("Group not found.")
+            self.stdout.write("Error: group not found")
             return
 
         users = []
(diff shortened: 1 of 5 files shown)


backend/gsr_booking/management/commands/change_group.py
```diff
@@ -0,0 +1,76 @@
...
+            return
+
+        if not (group := Group.objects.filter(name=group).first()):
+            self.stdout.write("Error: group not found")
+            return
+
+        users = []
+        failed_users = []
+        for username in usernames:
+            user = User.objects.filter(username=username).first()
```
vcai122, line 46: you can do your fancy := here if u want

backend/gsr_booking/management/commands/change_group.py
```diff
@@ -0,0 +1,76 @@
...
+            else:
+                users.append(user)
+
+        if failed_users:
+            self.stdout.write("Error: users not found: " + ", ".join(failed_users))
+            return
+
+        if reset:
+            group.memberships.exclude(Q(user__in=users) | Q(user=group.owner)).delete()
+        if reset or not remove:
```
vcai122, line 58: hmm.. since it should only ever be reset or remove since it doesn't make sense for them both to be true, would it be better to have like a "--mode/action/option" optional flag instead and expect the user to use a string that needs to be one of "add", "reset", or "remove" (we can default to add)?

backend/gsr_booking/management/commands/change_group.py
```diff
@@ -0,0 +1,76 @@
...
+                    user=user,
+                    defaults={
+                        "accepted": True,
+                        "type": GroupMembership.MEMBER,
+                        "pennkey_allow": True,
+                    },
+                )
+        elif remove:
+            group.memberships.filter(user__in=users).delete()
+
```
vcai122, line 70: should we just write grouped modified successfully or something since this doesn't the success message doesn't actually give extra information besides that it succeeded? or if we do the above we can insert the operation with fstring?

backend/gsr_booking/management/commands/get_reservations.py
```diff
@@ -0,0 +1,90 @@
...
+        if start and not (start := self.__convert_date(start)):
+            self.stdout.write("Error: invalid start date format")
+            return
+        if end and not (end := self.__convert_date(end)):
+            self.stdout.write("Error: invalid end date format")
+            return
+
+        reservation_filter = Q()
+        if current:
+            reservation_filter &= Q(start__lte=datetime.now()) & Q(end__gte=datetime.now())
```
vcai122, line 53: doesn't matter but you don't neede &= here

backend/gsr_booking/management/commands/get_reservations.py
```diff
@@ -0,0 +1,90 @@
...
+        time = kwargs["time"]
+
+        if start and not (start := self.__convert_date(start)):
+            self.stdout.write("Error: invalid start date format")
+            return
+        if end and not (end := self.__convert_date(end)):
+            self.stdout.write("Error: invalid end date format")
+            return
+
+        reservation_filter = Q()
```
vcai122, line 51: can also move this to the else block if you want

backend/gsr_booking/management/commands/get_reservations.py
```diff
@@ -0,0 +1,90 @@
...
+        if current:
+            reservation_filter &= Q(start__lte=datetime.now()) & Q(end__gte=datetime.now())
+        else:
+            if start:
+                reservation_filter &= Q(start__gte=start)
+            if end:
+                reservation_filter &= Q(end__lte=end)
+
```
vcai122, line 59: i don't think current and start/end needs to be mutually exclusive. it might make sense that to look for bookings that started after a certain time but is still going on. the reverse going on now but ended before some time doesn't really make sense to ask... but in any case the logic for both cases can still be achieved (the second would just give 0 bookings) by having default start to be lte now and end gte now, but overrided if start/end are passed, and then having the overall filter to just be the & of these.

vcai122 (approved):

backend/gsr_booking/management/commands/get_reservations.py
tuneerroy: ok ya that's a fair point

backend/gsr_booking/management/commands/change_group.py
tuneerroy: yeah that's probably a better approach. will change it

PR #184, opened 2023-04-15
Capitalize Fitness Room

vcai122:


backend/penndata/management/commands/get_fitness_snapshot.py
```diff
@@ -8,6 +8,10 @@
 from penndata.models import FitnessRoom, FitnessSnapshot
 
 
+def cap_string(s):
+    return " ".join([word[0].upper() + word[1:] for word in s.split(" ")])
```
tuneerroy, line 12: `split()` uses whitespace as its default separator so no need to specify space

backend/penndata/management/commands/rename_fitness_room.py
```diff
@@ -0,0 +1,24 @@
+from django.core.management.base import BaseCommand
+
+from penndata.models import FitnessRoom
+
+
+class Command(BaseCommand):
+    help = "Renames a fitness room."
```
tuneerroy, line 7: Instead of having this command take in inputs, just make it apply your `cap_string` function to all the FitnessRoom objects. It's easier to deal with since we just want to run this command once and then delete it. The following method will do that:
`FitnessRoom.objects.update(name=models.functions.Func('name', function=cap_string))`

tuneerroy (changes requested):
you're popping off Vincent :D

judtinzhang (approved):

tuneerroy (approved):
LGTM! Lmk when you squash and merge this so we can run the command. :)

judtinzhang:
Dang Tanner using the LGTM that's when you know

PR #186, opened 2023-04-20
Bugfix: Account for when User Populations can be null

judtinzhang:


backend/portal/logic.py
```diff
@@ -35,25 +35,32 @@ def get_user_populations(user):
...
+            )
+        ]
+        if user_info["student"]["graduation_year"]
+        else []
+    )
 
     school = [
         TargetPopulation.objects.get(kind=TargetPopulation.KIND_SCHOOL, population=x["name"])
         for x in user_info["student"]["school"]
+        if user_info["student"]["school"]
```
tuneerroy, line 51: this won't really do anything. I think you want something similar to how you handle `year`. Basically, for each `x`, it is going to check if `user_info["student"]["school"]` exists which we know to be `True` if it is already iterating. Similarly, it'll still throw an error if it doesn't exist.

backend/portal/logic.py
```diff
@@ -35,25 +35,32 @@ def get_user_populations(user):
...
     ]
 
     major = [
         TargetPopulation.objects.get(kind=TargetPopulation.KIND_MAJOR, population=x["name"])
         for x in user_info["student"]["major"]
+        if user_info["student"]["major"]
```
tuneerroy, line 57: same as above

backend/portal/logic.py
```diff
@@ -35,25 +35,32 @@ def get_user_populations(user):
...
     ]
 
     degree = [
         TargetPopulation.objects.get(kind=TargetPopulation.KIND_DEGREE, population=x["degree_type"])
         for x in user_info["student"]["major"]
+        if user_info["student"]["major"]
```
tuneerroy, line 63: same as above

tuneerroy (approved):
LGTM!

PR #187, opened 2023-05-14
User Fitness Preferences

JHawk0224:
Please review this, I have no idea what I'm doing so I probably missed something. This should just add user preferences for which fitness locations they favorited, and it should be similar/identical to the way laundry works.

I also tried to test it out, and while it runs, I couldn't figure out how (or became too lazy) to make a test user that can get authentication to test the route.

### DIFF
--- a/backend/penndata/views.py
+++ b/backend/penndata/views.py
@@ -429,10 +429,6 @@ class FitnessPreferences(APIView):
 
     def get(self, request):
 
-        # check fitness preferences existence first
-        if not hasattr(request.user.profile, "fitness_preferences"):
-            return Response({"rooms": []})
-
         preferences = request.user.profile.fitness_preferences.all()
 
         # returns all ids in a person's preferences
--- a/backend/penndata/views.py
+++ b/backend/penndata/views.py
@@ -416,7 +416,41 @@ class FitnessUsage(APIView):
                 "usage": {i: amt for i, amt in enumerate(usage_per_hour)},
             }
         )
+    
+class FitnessPreferences(APIView):
+    """
+    GET: returns list of a User's fitness preferences
 
+    POST: updates User fitness preferences by clearing past preferences
+    and resetting them with request data
+    """
+
+    permission_classes = [IsAuthenticated]
+
+    def get(self, request):
+
+        preferences = request.user.profile.fitness_preferences.all()
+
+        # returns all ids in a person's preferences
+        return Response({"rooms": preferences.values_list("id", flat=True)})
+
+    def post(self, request):
+
+        profile = request.user.profile
+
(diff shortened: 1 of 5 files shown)


backend/penndata/views.py
```diff
@@ -416,7 +416,41 @@ def get(self, request, room_id):
...
+
+        profile = request.user.profile
+
+        # clears all previous preferences in many-to-many
+        profile.fitness_preferences.clear()
+
+        ids = request.data["rooms"]
+
+        for id in ids:
+            hall = get_object_or_404(FitnessRoom, id=int(id))
```
tuneerroy, line 447: if a single `id` is invalid, then this will 404 after clearing the profile preferences. Before clearing the preferences, you should first get all the halls (use list comprehension) using the `get_object_or_404` which will avoid the problem.

backend/penndata/views.py
```diff
@@ -416,7 +416,41 @@ def get(self, request, room_id):
...
+        return Response({"rooms": preferences.values_list("id", flat=True)})
+
+    def post(self, request):
+
+        profile = request.user.profile
+
+        # clears all previous preferences in many-to-many
+        profile.fitness_preferences.clear()
+
+        ids = request.data["rooms"]
```
tuneerroy, line 444: make sure to check that the parameter is passed in before using it. a simple way to do this would be to do `request.data.get('rooms').... if ids is None: return <error stuff>`, but you can also first check `if rooms not in request.data: do <error stuff>`

backend/penndata/views.py
```diff
@@ -416,7 +416,41 @@ def get(self, request, room_id):
...
+
+    POST: updates User fitness preferences by clearing past preferences
+    and resetting them with request data
+    """
+
+    permission_classes = [IsAuthenticated]
+
+    def get(self, request):
+
+        preferences = request.user.profile.fitness_preferences.all()
```
tuneerroy, line 432: hm is this guaranteed to exist?

tuneerroy (changes requested):
Looks pretty good so far! 👍 Check out some of my comments below, make some changes, and then re-request a review (also tack on Vincent if you can).

JHawk0224:
I made the changes you said above, but I thought I should mention that most of this code was just copied from the laundry preferences code, so that should have the same issues.

backend/penndata/views.py
judtinzhang: Yes! According to the `Profile` model in users. It will be empty upon the creation of a user profile

tuneerroy (approved):
LGTM! Make sure to squash and merge when merging into `master`. Nicely done on your first PR! :D

PR #188, opened 2023-05-21
Remove Deprecated Codecov

tuneerroy:
Removes deprecated `codecov` package from backend environment. Disables the CI/CD steps that involve the package.


joyliu-q (approved):
We'll get to the bottom of this CI/CD step, but in the meanwhile, lgtm!

PR #190, opened 2023-05-22
Fix Date Parsing on Fitness Cron Job

vcai122:


tuneerroy (approved):
LGTM!

judtinzhang:
Since we aren't sure of the format that they are logging, I think it might be worthwhile to try-catch parsing the datetime under both formats, and then only propagating the one that works into the rest of the code.

judtinzhang (commented):
Comment above ^ lemme know what you guys think @tuneerroy @vcai122

vcai122:
@judtinzhang yea i also thought of this. i think both could be good. alternatively would something like python-dateutil be overkill for this? python-dateutil can sort of guess the time from lots of formats

judtinzhang:
@vcai122 oh wow that library looks really cool. I don't think it is overkill especially since we have no control on the format of that data day-to-day. I think thats a good idea :)

judtinzhang (approved):
Looks great! Thanks for doing this :)

PR #198, opened 2023-09-10
Add APIError catch to Dining

judtinzhang:


backend/dining/views.py
```diff
@@ -23,7 +23,10 @@ class Venues(APIView):
     """
 
     def get(self, request):
-        return Response(d.get_venues())
+        try:
+            return Response(d.get_venues())
+        except APIError as e:
+            return Response({"error": str(e)}, status=400)
```
vcai122, line 30: is this something we have to do for a lot of views?

judtinzhang: yea, you're right this seems like some behavior we would like to create a wrapper for. @vcai122 can you log this as a github issue? we can tackle it another time

vcai122: done

vcai122 (approved):

PR #202, opened 2023-09-13
Add Image to Fitness

vcai122:


backend/penndata/views.py
```diff
@@ -283,6 +283,9 @@ def get(self, request):
                 datetime.time(hour=int(hours), minute=int((hours % 1) * 60))
                 for _, hours in self.open_times.values()
             ]
+            # TODO: either unify naming OR add a field to model to indicate which image to use
```
vcai122, line 286: should we have this as db field? didn't do this to begin with bc i know it would involve someone manually filling in on admin page

judtinzhang: Yea I would personally like the db field approach more + making a script to load it in (something like `load_gsrs.py`). that way we can avoid the manual filling

judtinzhang (changes requested):
Glad that images are in! Would suggest db approach

Also would suggest rename PR :)

backend/penndata/models.py
```diff
@@ -30,6 +30,7 @@ def __str__(self):
 
 class FitnessRoom(models.Model):
     name = models.CharField(max_length=255)
+    image_url = models.URLField()
     # TODO "portal/images" does not exist, add back images
```
judtinzhang, line 34: Please remove these comments as they have been resolved

judtinzhang (approved):
One minor comment, but LGTM! ty :)

PR #205, opened 2023-09-17
Add Image to Fitness Admin and Portal Admin

vcai122:


backend/penndata/admin.py
```diff
@@ -1,10 +1,21 @@
...
 
 from penndata.models import AnalyticsEvent, Event, FitnessRoom, FitnessSnapshot, HomePageOrder
 
 
+# custom fitness room admin
+class FitnessRoomAdmin(admin.ModelAdmin):
+    def image_tag(self, instance):
+        return mark_safe('<img src="%s" height="300" />' % escape(instance.image_url))
+
+    image_tag.short_description = "Image"
```
judtinzhang, line 12: what do these fields do?

judtinzhang (approved):
LGTM thanks!

backend/penndata/admin.py
vcai122: short_description is like the label on the left on the admin portal. i removed the other one since mark_safe did the same thing. basically lets u render html

judtinzhang (approved):
So long as this runs locally and works for you, LGTM! thanks :D

backend/penndata/admin.py
judtinzhang: got it thanks!

PR #206, opened 2023-09-23
Target Populations Cronjob Every August

judtinzhang:
- Populates new target populations every August (particularly important for adding new graduation year)


tuneerroy (approved):
LGTM! We're sure that we only have to run this in the beginning of August, right?

judtinzhang:
I think that is what the cronjob should be doing. I double checked with the npm library we are using and it looks fine. Only true way is to see in August when I'm gone. But @vcai122 @tuneerroy an important thing is to ensure that for next year the class of 2028 is correctly in the `TargetPopulations` model

vcai122 (approved):

PR #207, opened 2023-09-26
Cache Laundry Routes

tuneerroy:
- [x] Cache Laundry


backend/laundry/views.py
```diff
@@ -152,29 +153,37 @@ class Preferences(APIView):
...
-            hall = get_object_or_404(LaundryRoom, hall_id=int(hall_id))
-            # adds all of the preferences given by the request
-            profile.laundry_preferences.add(hall)
+        # clears all previous preferences in many-to-many
+        preferences.clear()
+        preferences.add(*halls)
 
-        profile.save()
+        # clear cache
+        cache.delete(key)
```
judtinzhang, line 186: Does this ever error if the key isn't present (which would happen for the first time someone adds a preference)? If so might be worth to check

backend/utils/cache.py
```diff
@@ -0,0 +1,5 @@
+MINUTE_IN_SECONDS = 60
```
judtinzhang, line 1: I think maybe an `IntEnum` might work better here? Would be easier to import to other files and also syntax-wise we can do `Cache.SECONDS` or like `Cache.MINUTE` or `30 * Cache.SECONDS`

judtinzhang (commented):
Looks good! Some minor questions + comments

backend/laundry/urls.py
```diff
@@ -1,13 +1,19 @@
...
-    path("status/", Status.as_view(), name="status"),
+    path("hall/<hall_id>/", cache_page(MINUTE_IN_SECONDS)(HallInfo), name="hall-info",),
+    path("usage/<hall_id>/", cache_page(MINUTE_IN_SECONDS)(HallUsage), name="hall-usage",),
+    path(
+        "rooms/<hall_ids>",
+        cache_page(MINUTE_IN_SECONDS)(MultipleHallInfo),
+        name="multiple-hall-info",
+    ),
+    path("halls/ids/", cache_page(MONTH_IN_SECONDS_APPROX)(Ids), name="hall-ids",),
+    path("status/", cache_page(30)(Status), name="status",),
```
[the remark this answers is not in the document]
tuneerroy: I think this route is hit everytime someone opens up laundry on the app. If so, I think it makes sense for it to be briefly cached, but I'll ask Jordan about it.

backend/laundry/views.py
tuneerroy: Nope! returns a boolean saying whether or not it actually gets deleted.

backend/utils/cache.py
tuneerroy: sgtm!

backend/laundry/urls.py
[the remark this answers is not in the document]
judtinzhang: To my knowledge I remember that iOS caches these kind of things on their side. And if they don't, I think design-wise it would be best of this kind of caching is done on frontend side just because this route should be precise and a direct representation of what the laundry system is

backend/laundry/views.py
```diff
@@ -152,29 +153,37 @@ class Preferences(APIView):
     """
 
     permission_classes = [IsAuthenticated]
+    key = "laundry_preferences:{{user_id}}"
```
judtinzhang, line 156: Oh this is cool, could you comment below the doc for this syntax?

backend/utils/cache.py
```diff
@@ -0,0 +1,9 @@
+from enum import IntEnum
+
+
+class Cache(IntEnum):
+    MINUTE = 60
```
judtinzhang, line 5: nit: can we standardize the format of `(multiplier)  * ENUM` or something

judtinzhang (approved):
LGTM! Just some nits. There is a django-check error so once that is good feel free to merge :) thanks for this

PR #210, opened 2023-10-02
Command to Reset Wharton Caching for GSRs

vcai122:


backend/gsr_booking/management/commands/reset_wharton_caching.py
```diff
@@ -0,0 +1,27 @@
+from django.core.management.base import BaseCommand
+
+from gsr_booking.models import GroupMembership
+
+
+groups = [
```
vcai122, line 6: these numbers are how many group memberships start with those letters

judtinzhang (approved):

PR #219, opened 2023-10-03
Create Models for Subletting

dr-Jess:
A few notes:
 - Some stuff in portal is commented out bc it was destroying everything else, should be findable with a quick search for TODO
 - I think the way I did URLS is throwing major errors, so I'd doubly appreciate a look at that
 - Models probably has some mistakes, but we put what made sense for now
 - Version history is let's say not optimally done, let's talk about this at the next meeting


backend/subletting/__init__.py
judtinzhang, line 1: Thanks for creating the django app! Please also add this app into `INSTALLED_APPS` in our `base.py`

backend/portal/serializers.py
```diff
@@ -31,12 +31,14 @@ class Meta:
     def create(self, validated_data):
         club_code = validated_data["club_code"]
         # ensures user is part of club
+        '''TODO: Uncomment this, commented while testing
```
judtinzhang, line 34: Please leaves these changes unstaged before pushing. That way it won't show on the final code, but you'll still have it in your local setup

backend/portal/models.py
```diff
@@ -65,7 +65,7 @@ class PollVote(models.Model):
     poll = models.ForeignKey(Poll, on_delete=models.CASCADE)
     poll_options = models.ManyToManyField(PollOption)
     created_date = models.DateTimeField(default=timezone.now)
-    target_populations = models.ManyToManyField(TargetPopulation, blank=True)
+    target_Fpopulations = models.ManyToManyField(TargetPopulation, blank=True)
```
judtinzhang, line 68: I think typo here?

backend/subletting/admin.py
```diff
@@ -0,0 +1,8 @@
+from django.contrib import admin
+
+from subletting.models import Offer, Amenity, Sublet, Image
+
+admin.site.register(Offer)
+admin.site.register(Amenity)
+admin.site.register(Sublet)
+admin.site.register((Image))
```
judtinzhang, line 8: nit: would probably want to remove double paranthesis on `Image`

backend/subletting/models.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+User = get_user_model()
+
+class Offer(models.Model):
+    user = models.ForeignKey(User, on_delete=models.CASCADE, related_name="offers_made")
+    sublet = models.ForeignKey("Sublet", on_delete=models.CASCADE, related_name="offers")
+    #TODO: check if the way we're handling phone/email is fine, validate on frontend/routes that at least one is not null
+    email = models.EmailField(max_length=255, null=True)
+    phone_number = models.CharField(max_length=20, null=True)
```
judtinzhang, line 13: Validation here might be tricky. We can either
1. Validate phone number in serializer
2. Import a library to do it
For simplicity I think option 1 is fine

backend/subletting/models.py
```diff
@@ -0,0 +1,53 @@
...
+    baths = models.IntegerField()
+    description = models.TextField()
+    external_link = models.CharField(max_length=255, null=True)
+    min_price = models.IntegerField()
+    max_price = models.IntegerField()
+    created_at = models.DateTimeField(auto_now_add=True)
+    #TODO: would a last updated datetime be helpful?
+    close_at = models.DateTimeField()
+    #TODO: do we need time or just date on these starts?
+    sublet_start_date = models.DateField()
```
judtinzhang, line 43: nit: suggest removing the `sublet` prefix because we are under the `Sublet` class

backend/subletting/models.py
```diff
@@ -0,0 +1,53 @@
...
+    sublet_end_date = models.DateField()
+
+    def __str__(self):
+        return f"{self.title} by {self.subletter}"
+
+#Not sure how exactly things are handled S3-side, please review
+class Image(models.Model):
```
judtinzhang, line 50: nit: suggest renaming this to `SubletImage` or something to not confuse it with a generic image

backend/subletting/models.py
```diff
@@ -0,0 +1,53 @@
...
+    address = models.CharField(max_length=255)
+    beds = models.IntegerField()
+    baths = models.IntegerField()
+    description = models.TextField()
+    external_link = models.CharField(max_length=255, null=True)
+    min_price = models.IntegerField()
+    max_price = models.IntegerField()
+    created_at = models.DateTimeField(auto_now_add=True)
+    #TODO: would a last updated datetime be helpful?
+    close_at = models.DateTimeField()
```
judtinzhang, line 41: nit: suggest renaming to `expires_at` and make the `DateTimeField` because I would imagine people would want a specific time to close

backend/subletting/models.py
```diff
@@ -0,0 +1,53 @@
...
+    subletter = models.ForeignKey(User, on_delete=models.CASCADE)
+    sublettees = models.ManyToManyField(User, through=Offer, related_name="sublets_offered")
+    amenities = models.ManyToManyField(Amenity)  # TODO: not sure if anything else should go into this as params
+
+    title = models.CharField(max_length=255)
+    address = models.CharField(max_length=255)
+    beds = models.IntegerField()
+    baths = models.IntegerField()
+    description = models.TextField()
+    external_link = models.CharField(max_length=255, null=True)
```
judtinzhang, line 36: I think we can use a `URLField` here

backend/subletting/models.py
```diff
@@ -0,0 +1,53 @@
...
+class Sublet(models.Model):
+    subletter = models.ForeignKey(User, on_delete=models.CASCADE)
+    sublettees = models.ManyToManyField(User, through=Offer, related_name="sublets_offered")
+    amenities = models.ManyToManyField(Amenity)  # TODO: not sure if anything else should go into this as params
+
+    title = models.CharField(max_length=255)
+    address = models.CharField(max_length=255)
+    beds = models.IntegerField()
+    baths = models.IntegerField()
+    description = models.TextField()
```
judtinzhang, line 35: I think we can make this nullable / blank

backend/subletting/models.py
```diff
@@ -0,0 +1,53 @@
...
+class Amenity(models.Model):
+    name = models.CharField(max_length=100, primary_key=True)
+
+    def __str__(self):
+        return self.name
+
+class Sublet(models.Model):
+    subletter = models.ForeignKey(User, on_delete=models.CASCADE)
+    sublettees = models.ManyToManyField(User, through=Offer, related_name="sublets_offered")
+    amenities = models.ManyToManyField(Amenity)  # TODO: not sure if anything else should go into this as params
```
judtinzhang, line 29: I'm not sure if it should be required to list amenities for a Sublet, so we should also add `blank=True` here

backend/subletting/models.py
```diff
@@ -0,0 +1,53 @@
...
+
+class Amenity(models.Model):
+    name = models.CharField(max_length=100, primary_key=True)
+
+    def __str__(self):
+        return self.name
+
+class Sublet(models.Model):
+    subletter = models.ForeignKey(User, on_delete=models.CASCADE)
+    sublettees = models.ManyToManyField(User, through=Offer, related_name="sublets_offered")
```
judtinzhang, line 28: This is great, thanks! I think we should also make this `blank=True` too because initially during `Sublet` creation this will (and should) be empty

judtinzhang (commented):
Great work you both! Thanks for getting this done -- I left a couple of minor comments about the schema. So far looks great!

backend/subletting/urls.py
```diff
@@ -0,0 +1,9 @@
+from django.urls import path
+from rest_framework import routers
+
+
+#from subletting.views import{}
+
+app_name = "subletting"
+
+routers = routers.DefaultRouter()
```
judtinzhang, line 9: I'd also imagine the URLs is throwing an error because you haven't defined `urlpatterns` (see other apps for this)

backend/subletting/models.py
vcai122, line 13: I think we should also do blank=True, so if someone is editing on admin portal it won't get mad for leaving one empty

vcai122: any reason not to use a PhoneNumberField like you suggested?

backend/subletting/models.py
vcai122: @judtinzhang on a similar vein should we allow saving a "draft", i.e. maybe adding a boolean field like is_draft? although this could also mess with query speed...

backend/subletting/models.py
```diff
@@ -0,0 +1,53 @@
+from django.contrib.auth import get_user_model
+from django.db import models
+from django.utils import timezone
+
+
+User = get_user_model()
+
+class Offer(models.Model):
```
vcai122, line 8: I think I said that having this be a "through" model would automatically make it have composite key (User, Sublet), turns out I lied. So it might be worth having unique together constraint

vcai122 (commented):
Looks pretty good so far :D

backend/subletting/models.py
judtinzhang: @vcai122 I don't think django has this field by default. We would need to import

backend/subletting/models.py
judtinzhang: I like that idea, like having a "Save for later" feature. Perhaps we can do something similar to Polls / Posts (e.g. with an enum)

backend/subletting/models.py
dr-Jess: @judtinzhang just wanted to check, I agree with the rename but it seems to already be a DateTimeField() unless I'm misunderstanding something?

backend/subletting/models.py
dr-Jess: I've added blank for sublettees; I assume that the subletter shouldn't be blank?

backend/sublet/urls.py
```diff
@@ -0,0 +1,26 @@
...
+additional_urls = [
+    path("amenities/", Amenities.as_view(), name="amenities"),
+    path("favorites/", UserFavorites.as_view(), name="user-favorites"),
+    path(
+        "properties/<sublet_id>/favorites/",
+        Favorites.as_view({"post": "create", "delete": "destroy"}),
+    ),
+    path(
+        "properties/<sublet_id>/offers/",
+        Offers.as_view({"get": "list", "post": "create", "delete": "destroy"}),
```
vcai122, line 22: is there a reason for calling it list, create, and destroy in the views and then renaming it here?

backend/sublet/views.py
```diff
@@ -0,0 +1,197 @@
...
+        serializer = SubletSerializer(queryset, many=True)
+        return Response(serializer.data)
+
+    @action(detail=True, methods=["get"])
+    def view_property(self, request, pk=None):
+        """Returns details of a specific Sublet."""
+        sublet = self.get_object()
+        serializer = SubletSerializer(sublet)
+        return Response(serializer.data)
+
```
vcai122, line 138: Can you not do the single view and the list view in the same get function?

backend/sublet/views.py
```diff
@@ -0,0 +1,197 @@
...
+        if ends_after:
+            queryset = queryset.filter(end_date__gt=ends_after)
+        if min_price:
+            queryset = queryset.filter(min_price__gte=min_price)
+        if max_price:
+            queryset = queryset.filter(max_price__lte=max_price)
+        if beds:
+            queryset = queryset.filter(beds=beds)
+        if baths:
+            queryset = queryset.filter(baths=baths)
```
vcai122, line 126: Not sure if Justin/Tuneer would agree, but we could have the query params be literally the db query names so start_date__lt instead of starts_before, so we could literally just pass in **request.query_params

backend/sublet/views.py
```diff
@@ -0,0 +1,197 @@
...
+        return Response(serializer.data)
+
+    @action(detail=True, methods=["get"])
+    def view_property(self, request, pk=None):
+        """Returns details of a specific Sublet."""
+        sublet = self.get_object()
+        serializer = SubletSerializer(sublet)
+        return Response(serializer.data)
+
+
```
vcai122, line 139: You probably don't need to manually implement these functions / make the calls to the serializer. 
If you set the queryset to a user's favorite objects, Django should know how to handle it for you.

We can explain more Sunday!

Same with Offers

vcai122 (changes requested):
Left some comments, good work! We can go over more in detail on Sunday

backend/sublet/views.py
```diff
@@ -0,0 +1,197 @@
...
+    Returns a list of all offers for the sublet matching the provided ID.
+
+    create:
+    Create an offer on the sublet matching the provided ID.
+
+    destroy:
+    Delete the offer between the user and the sublet matching the ID.
+    """
+
+    # TODO: implement permissions
```
judtinzhang, line 176: Agreed! Need permissions here but otherwise looks complete thanks

backend/sublet/serializers.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def create(self, validated_data):
+        validated_data["user"] = self.context["request"].user
+        return super().create(validated_data)
+
+
+class SubletSerializer(serializers.ModelSerializer):
+    amenities = AmenitySerializer(many=True, required=False)
+    #favorites = FavoriteSerializer(many=True, required=False, read_only=True)
+    sublettees = OfferSerializer(many=True, required=False, read_only=True)
```
judtinzhang, line 32: I think this will error, same as favorites. Reason why is because `Offers` and `Favorites` are not sub-components of the `Sublet` class, so it is pretty non-trivial to nest these serializers like that. I suggest removing both favorites and offers here and have the frontend just use the `<sublet_id>/offers` and `<sublet_id>/favorites` route to get this info

judtinzhang (commented):
Looks in great shape! Almost done :D

backend/sublet/views.py
```diff
@@ -0,0 +1,197 @@
...
+        sublet.save()
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+    @action(detail=False, methods=["get"])
+    def browse(self, request):
+        """Returns a list of Sublets that match query parameters and user ownership."""
+        # Get query parameters from request (e.g., amenities, user_owned)
+        amenities = request.query_params.getlist("amenities")
+        subletter = request.query_params.get(
+            "subletter", False
```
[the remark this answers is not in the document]
dr-Jess: Fixed, for now.

backend/sublet/views.py
dr-Jess: Done :)

backend/sublet/serializers.py
dr-Jess: I left amenities in, but removed sublettees here since it's not useful.

backend/sublet/views.py
tuneerroy: don't think we need to worry about SQL injection attacks since Django will protect against that in the `filter` clause. but from the frontend side, it is probably nier for them to be declaring stuff as "max_price" rather than "max_price__lte" so this is fine imo.

backend/sublet/serializers.py
```diff
@@ -0,0 +1,93 @@
...
+        model = Sublet
+        exclude = ["favorites"]
+        read_only_fields = ["id", "created_at", "subletter", "sublettees"]
+
+    def parse_amenities(self, raw_amenities):
+        if isinstance(raw_amenities, list):
+            ids = raw_amenities
+        else:
+            ids = (
+                list() if len(raw_amenities) == 0 else [str(id) for id in raw_amenities.split(",")]
```
vcai122, line 38: why do `str(id)`? 
I guess my confusion is if you're doing .split, raw_amenities is already a string? did you mean to do int/did this work when u tested it?

vcai122 (commented):
Just took a quick glance, but its looking good. When you wrap this up make sure to remove the commented out code you had for testing.

backend/sublet/serializers.py
judtinzhang: It works. To be honest I'm not exactly sure how this works but this is a django `QueryDict` which is part of `request.POST` (not `request.data`). Very bad syntax, we should probably change, but this is an alternative way to add amenities to the sublet

backend/sublet/views.py
```diff
@@ -0,0 +1,208 @@
...
+        user = self.request.user
+        return user.sublets_favorited
+
+    def create(self, request, *args, **kwargs):
+        sublet_id = int(self.kwargs["sublet_id"])
+        queryset = self.get_queryset()
+        if queryset.filter(id=sublet_id).exists():
+            raise exceptions.NotAcceptable("Favorite already exists")
+        sublet = get_object_or_404(Sublet, id=sublet_id)
+        self.get_queryset().add(sublet)
```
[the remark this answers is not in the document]
dr-Jess: Passes in the tests I just pushed, so I think it's probably fine.

backend/sublet/models.py
```diff
@@ -0,0 +1,65 @@
...
+
+class Offer(models.Model):
+    class Meta:
+        constraints = [models.UniqueConstraint(fields=["user", "sublet"], name="unique_offer")]
+
+    user = models.ForeignKey(User, on_delete=models.CASCADE, related_name="offers_made")
+    sublet = models.ForeignKey("Sublet", on_delete=models.CASCADE, related_name="offers")
+    # TODO: Make sure phone_number is being validated by serializers/frontend
+    email = models.EmailField(max_length=255, null=True, blank=True)
+    phone_number = models.CharField(max_length=20, null=True, blank=True)
```
judtinzhang, line 16: I think we would need to validate the phone number somehow, please make sure to either make this the PhoneNumberField or check in serializer update and create functions

backend/sublet/models.py
```diff
@@ -0,0 +1,74 @@
...
+    name = models.CharField(max_length=255, primary_key=True)
+
+    def __str__(self):
+        return self.name
+
+
+class Sublet(models.Model):
+    subletter = models.ForeignKey(User, on_delete=models.CASCADE)
+    sublettees = models.ManyToManyField(
+        User, through=Offer, related_name="sublets_offered", null=True, blank=True
```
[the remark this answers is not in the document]
judtinzhang: nit: reminder for this and other `null=True` additions

backend/sublet/serializers.py
```diff
@@ -0,0 +1,93 @@
...
+    class Meta:
+        model = Amenity
+        fields = "__all__"
+
+
+class OfferSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Offer
+        fields = "__all__"
+        read_only_fields = ["id", "created_date", "user"]
```
judtinzhang, line 16: I think the field in models is `created_at`, not `created_date`

backend/sublet/serializers.py
```diff
@@ -0,0 +1,93 @@
...
+            self.context["request"].user == instance.subletter
+            or self.context["request"].user.is_superuser
+        ):
+            instance.delete()
+        else:
+            raise serializers.ValidationError("You do not have permission to delete this sublet.")
+
+
+# simple sublet serializer for use when pulling all serializers/etc
+class SimpleSubletSerializer(serializers.ModelSerializer):
```
judtinzhang, line 81: I think we would need to attach `image_url` here

backend/sublet/views.py
```diff
@@ -0,0 +1,208 @@
...
+            "subletter", "false"
+        )  # Defaults to False if not specified
+        starts_before = request.query_params.get("starts_before", None)
+        starts_after = request.query_params.get("starts_after", None)
+        ends_before = request.query_params.get("ends_before", None)
+        ends_after = request.query_params.get("ends_after", None)
+        min_price = request.query_params.get("min_price", None)
+        max_price = request.query_params.get("max_price", None)
+        beds = request.query_params.get("beds", None)
+        baths = request.query_params.get("baths", None)
```
tuneerroy, line 113: nit: if we're going to be doing `request.query_params` can we just extract that to a separate variable

backend/sublet/views.py
```diff
@@ -0,0 +1,208 @@
...
+        queryset = self.get_queryset()
+        if queryset.filter(id=sublet_id).exists():
+            raise exceptions.NotAcceptable("Favorite already exists")
+        sublet = get_object_or_404(Sublet, id=sublet_id)
+        self.get_queryset().add(sublet)
+        return Response(status=status.HTTP_201_CREATED)
+
+    def destroy(self, request, *args, **kwargs):
+        queryset = self.get_queryset()
+        sublet = get_object_or_404(queryset, pk=int(self.kwargs["sublet_id"]))
```
tuneerroy, line 164: kinda curious if this works without casting the id to int first, can we try it out?

backend/sublet/migrations/0003_alter_offer_phone_number.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 3.2.22 on 2023-10-29 17:24
+
+import phonenumber_field.modelfields
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
judtinzhang, line 7: nit: let's delete all these migration files and re-makemigrations to put all of these changes into one migration file

backend/sublet/models.py
```diff
@@ -0,0 +1,64 @@
...
+User = get_user_model()
+
+
+class Offer(models.Model):
+    class Meta:
+        constraints = [models.UniqueConstraint(fields=["user", "sublet"], name="unique_offer")]
+
+    user = models.ForeignKey(User, on_delete=models.CASCADE, related_name="offers_made")
+    sublet = models.ForeignKey("Sublet", on_delete=models.CASCADE, related_name="offers")
+    # TODO: Make sure phone_number is being validated by serializers/frontend
```
judtinzhang, line 15: nit: let's remove these comments that have been resolved

backend/sublet/views.py
```diff
@@ -0,0 +1,213 @@
...
+        return Response(response_data)
+
+
+class UserFavorites(generics.ListAPIView):
+    serializer_class = SimpleSubletSerializer
+    permission_classes = [IsAuthenticated]
+
+    def get_queryset(self):
+        user = self.request.user
+        # print(type(user.favorite_set))
```
judtinzhang, line 37: nit: let's remove these commented out test-code

backend/sublet/views.py
```diff
@@ -0,0 +1,213 @@
...
+
+        queryset = Sublet.objects.all().filter(expires_at__gte=timezone.now())
+
+        # Apply filters based on query parameters
+        if amenities:
+            queryset = queryset.filter(amenities__name__in=amenities)
+        if subletter.lower() == "true":
+            queryset = queryset.filter(subletter=request.user)
+        if starts_before:
+            queryset = queryset.filter(start_date__lt=starts_before)
```
judtinzhang, line 123: nit: can we test one of two of these if statements just to guarantee that they work? no need to test them all unless you want to

judtinzhang (approved):

vcai122 (approved):

PR #224, opened 2023-11-06
GSR Revamp

vcai122:
TODO: Fix test cases!


backend/gsr_booking/views.py
```diff
@@ -193,7 +191,7 @@ def get_queryset(self):
 
 class CheckWharton(APIView):
     def get(self, request):
-        return Response({"is_wharton": BW.is_wharton(request.user)})
+        return Response({"is_wharton": WhartonGSRBooker.is_wharton(request.user)})
```
judtinzhang, line 194: Let's add a check to see if `request.user` is not None first and default return false if that is the case. That way we can still let not-logged in users to hit this route.

backend/gsr_booking/api_wrapper.py
```diff
@@ -234,117 +132,37 @@ def get_availability(self, lid, start, end, username):
...
-            many=True,
-        ).data
-
-        for wharton_booking in wharton_bookings:
-            booking_ids.add(wharton_booking["booking_id"])
-
+        url = f"{WHARTON_URL}{user.username}/reservations"
+        bookings = self.request("GET", url).json()["bookings"]
+
+        bookings = [booking for booking in bookings if datetime.datetime.strptime(booking["end"], "%Y-%m-%dT%H:%M:%S%z") >= timezone.localtime()]
```
judtinzhang, line 140: Might be cleaner if we can have a `serialize_reservation` function here to prettify the output (like what you did a few lines down) and then we can consolidate the two list comprehensions into one. Or similarly you could just not use the function and directly consolidate.

backend/gsr_booking/api_wrapper.py
```diff
@@ -358,14 +176,15 @@ def update_token(self):
...
         response = requests.post(f"{API_URL}/1.1/oauth/token", body).json()
 
         if "error" in response:
             raise APIError(f"LibCal: {response['error']}, {response.get('error_description')}")
         self.expiration = timezone.localtime() + datetime.timedelta(seconds=response["expires_in"])
         self.token = response["access_token"]
-
-    def request(self, *args, **kwargs):
+    
+    def request(self, *arg, **kwargs):
```
judtinzhang, line 187: Hm, seems like you did this multiple times, any reason why we name to `arg`?

backend/gsr_booking/api_wrapper.py
```diff
@@ -489,38 +230,76 @@ def book_room(self, rid, start, end, user, test=False):
...
-        )
+    def cancel_room(self, booking_id, user):
+        """Cancels room"""
+        # gsr_booking = get_object_or_404(GSRBooking, booking_id=booking_id)
+        # if gsr_booking:
+        #     if user != gsr_booking.user and user != gsr_booking.reservation.creator:
+        #         raise APIError("Error: Unauthorized: This reservation was booked by someone else.")
+        #     gsr_booking.is_cancelled = True
+        #     gsr_booking.save()
+        response = self.request("POST", f"{API_URL}/1.1/space/cancel/{booking_id}").json()
```
judtinzhang, line 256: nit: check status code first

backend/gsr_booking/api_wrapper.py
```diff
@@ -489,38 +230,76 @@ def book_room(self, rid, start, end, user, test=False):
...
+            if end and not start == end:
+                range_str += "," + end
+        else:
+            start_datetime = None
+
+        # filters categories and then gets extra information on each room
+        
+        response = self.request("GET", f"{API_URL}/1.1/space/category/{gid}").json()
+        items = response[0]["items"]
+        items = ",".join([str(x) for x in items])
```
judtinzhang, line 280: nit: tech debt haha pls rename `x` variable for clarity

backend/gsr_booking/api_wrapper.py
```diff
@@ -489,38 +230,76 @@ def book_room(self, rid, start, end, user, test=False):
...
+                del room["formid"]
+            # enforce date filter
+            # API returns dates outside of the range, fix this manually
+            room["availability"] = [
+                {
+                    'start_time': time['from'],
+                    'end_time': time['to']
+                }
+                for time in room["availability"]
+                if not start_datetime or datetime.datetime.strptime(time["from"][:-6], "%Y-%m-%dT%H:%M:%S") >= start_datetime
```
judtinzhang, line 299: Is there a reason we did the `[:-6]`, like why couldn't we do the whole `time["from"]` and add an extra `%z` at the end of the date formatter string

backend/gsr_booking/api_wrapper.py
```diff
@@ -532,38 +311,175 @@ def get_affiliation(self, email):
...
 
-    def cancel_room(self, user, booking_id):
-        """Cancels room"""
-        gsr_booking = get_object_or_404(GSRBooking, booking_id=booking_id)
-        if gsr_booking:
-            if user != gsr_booking.user and user != gsr_booking.reservation.creator:
+            
+class BookingHandler:
+    def __init__(self, WBW=None, LBW=None):
+        self.WBW = WBW or WhartonBookingWrapper()
```
judtinzhang, line 317: Oh interesting does this work? Or would we need to do `WBW if WBW else WhartonBookingWrapper()`?

backend/gsr_booking/models.py
```diff
@@ -120,7 +121,7 @@ class Reservation(models.Model):
     start = models.DateTimeField(default=timezone.now)
     end = models.DateTimeField(default=timezone.now)
     creator = models.ForeignKey(User, on_delete=models.CASCADE)
-    group = models.ForeignKey(Group, on_delete=models.CASCADE)
+    group = models.ForeignKey(Group, on_delete=models.CASCADE, null=True, blank=True)
```
judtinzhang, line 124: I think you might need to make a migration change for this

backend/gsr_booking/api_wrapper.py
```diff
@@ -532,38 +311,175 @@ def get_affiliation(self, email):
...
+        self.LBW = LBW or LibCalBookingWrapper()
+
+
+    def format_members(self, members):
+        return [
```
judtinzhang, line 322: Can we have a variable like `PREFIX='user'__` and then instead of `6:` we could use `len(PREFIX)` for clarity

backend/gsr_booking/api_wrapper.py
```diff
@@ -532,38 +311,175 @@ def get_affiliation(self, email):
...
+            (User(**{ k[6:]: v for k, v in member.items() if k.startswith('user__') }), # temp user object
+            member['left'])
+            for member in members
+        ]
+
+    def get_wharton_members(self, group, gsr_id):
+        now = timezone.localtime()
+        ninty_min = datetime.timedelta(minutes=90)
+        zero_min = datetime.timedelta(minutes=0)
+        ret = GroupMembership.objects.filter(group=group, is_wharton=True).values('user').annotate(left=ninty_min-Coalesce(Sum(
```
judtinzhang, line 332: if we are removing this `is_wharton` cache, how would we change this code? perhaps we just keep the `is_wharton` field for Labs people?

backend/gsr_booking/api_wrapper.py
```diff
@@ -532,38 +311,175 @@ def get_affiliation(self, email):
...
+            F('user__gsrbooking__end')-F('user__gsrbooking__start'),
```
judtinzhang, line 333: Wow nice query! This `Coalesce` + filter with the annotation is really clean.

backend/gsr_booking/api_wrapper.py
```diff
@@ -532,38 +311,175 @@ def get_affiliation(self, email):
...
+            filter=Q(user__gsrbooking__gsr__gid=gsr_id) & Q(user__gsrbooking__is_cancelled=False) & Q(user__gsrbooking__end__gte=now)
+        ), zero_min)).filter(Q(left__gt=zero_min)).values('user__id', 'user__username', 'left').order_by('?')[:3]
```
judtinzhang, line 335: Any reason we only do the first 3? What if a Wharton labs member wants to book for longer than 90 minutes?

backend/gsr_booking/api_wrapper.py
```diff
@@ -532,38 +311,175 @@ def get_affiliation(self, email):
...
+
+    def get_libcal_members(self, group):
+        day_start = timezone.localtime().replace(hour=0, minute=0, second=0, microsecond=0)
```
judtinzhang, line 342: nit: is there a `.date()` provided in `timezone` so we can avoid the `replace`?

backend/gsr_booking/api_wrapper.py
```diff
@@ -532,38 +311,175 @@ def get_affiliation(self, email):
...
+        day_end = day_start + datetime.timedelta(days=1)
+        two_hours = datetime.timedelta(hours=2)
+        zero_min = datetime.timedelta(minutes=0)
+
```
judtinzhang, line 346: nit: for this and the wharton query, can we leave some sort of comments to justify the difference in filtering, such as wharton allows 90 at any given time for any given gid and libcal allows 120 on any given day

backend/gsr_booking/api_wrapper.py
```diff
@@ -532,38 +311,175 @@ def get_affiliation(self, email):
...
+        ret = GroupMembership.objects.filter(group=group).values('user').annotate(left=two_hours-Coalesce(Sum(
+            F('user__gsrbooking__end')-F('user__gsrbooking__start'),
+            filter=Q(user__gsrbooking__gsr__kind=GSR.KIND_LIBCAL) & Q(user__gsrbooking__is_cancelled=False) & Q(user__gsrbooking__start__gte=day_start) & Q(user__gsrbooking__end__lte=day_end)
+        ), zero_min)).filter(Q(left__gt=zero_min)).values('user__id', 'user__username', 'user__first_name', 'user__last_name', 'user__email', 'left').order_by('?')[:4]
```
judtinzhang, line 350: any reason for the `:4`?

backend/gsr_booking/api_wrapper.py
```diff
@@ -532,38 +311,175 @@ def get_affiliation(self, email):
...
+class BookingHandler:
+    def __init__(self, WBW=None, LBW=None):
+        self.WBW = WBW or WhartonBookingWrapper()
+        self.LBW = LBW or LibCalBookingWrapper()
+
+
+    def format_members(self, members):
+        return [
+            (User(**{ k[6:]: v for k, v in member.items() if k.startswith('user__') }), # temp user object
+            member['left'])
```
judtinzhang, line 324: nit: also perhaps rename `left` to maybe `credits` for clarity?

backend/gsr_booking/api_wrapper.py
```diff
@@ -532,38 +311,175 @@ def get_affiliation(self, email):
...
+
+                if (curr_start := curr_end) >= end:
+                    break
+        except APIError as e:
+            raise APIError(f"{str(e)}. Was only able to book {start.strftime('%H:%M')} - {curr_start.strftime('%H:%M')}")
+
+        return reservation
+    
+    def cancel_room(self, booking_id, user):
+        # I think this should prefetch reservation, reservation__gsrbooking_set, and gsr ?
```
judtinzhang, line 401: yep! looks correct to me

backend/gsr_booking/api_wrapper.py
```diff
@@ -532,38 +311,175 @@ def get_affiliation(self, email):
...
+        if gsr_booking := GSRBooking.objects.filter(booking_id=booking_id).prefetch_related(Prefetch('reservation__gsrbooking_set'), Prefetch('gsr')).first():
+            if gsr_booking.user != user and gsr_booking.reservation.creator != user:
                 raise APIError("Error: Unauthorized: This reservation was booked by someone else.")
+            
+            (self.WBW.cancel_room if gsr_booking.gsr.kind == GSR.KIND_WHARTON else self.LBW.cancel_room)(booking_id, gsr_booking.user)
```
judtinzhang, line 406: lol nice code haha

backend/gsr_booking/api_wrapper.py
```diff
@@ -532,38 +311,175 @@ def get_affiliation(self, email):
...
-            # accumulates total minutes over all reservations
-            total_minutes += int((reservation.end.timestamp() - reservation.start.timestamp()) / 60)
-        # 120 minutes per day
-        return {CreditType.LIBCAL.value: 120 - total_minutes}
+            reservation = gsr_booking.reservation
+            if all(booking.is_cancelled for booking in reservation.gsrbooking_set.all()):
+                reservation.is_cancelled = True
+                reservation.save()                
+        else:
+            try:
```
judtinzhang, line 416: Perhaps something like this? Nested try-excepts kinda look weird imo lol:
```
services = [self.WBW, self.LBW]

for service in services:
    try:
        service.cancel_room(booking_id, user)
        break
    except APIError:
        continue

raise APIError("Error: Unknown booking id")
```

backend/gsr_booking/api_wrapper.py
```diff
@@ -532,38 +311,175 @@ def get_affiliation(self, email):
...
+                except APIError:
+                    raise APIError("Error: Unknown booking id")
+                
+        
+    def get_availability(self, lid, gid, start, end, user, group=None):            
+        gsr = get_object_or_404(GSR, gid=gid)
+
+        # select a random user from the group if booking wharton gsr
+        if gsr.kind == GSR.KIND_WHARTON and group is not None:
+            wharton_members = group.memberships.filter(is_wharton=True)
```
judtinzhang, line 430: Hm, I feel like we don't need the additional computation for getting random. I think we can just always use the first person with (`.first()`). Maybe we would need this random so that wharton doesn't get suspicious when this one user is using this system a lot but otherwise I'm not sure

backend/gsr_booking/api_wrapper.py
```diff
@@ -532,38 +311,175 @@ def get_affiliation(self, email):
...
+                    data['room_name'] = f"[Me] {data['room_name']}"
+                else:
+                    data['room_name'] = f"[{group.name}] {data['room_name']}"
+                ret.append(data)
+        else:
+            ret = GSRBookingSerializer(bookings, many=True).data
+        
+        # deal with bookings made directly through wharton (not us)
+        try:
+            wharton_bookings = self.WBW.get_reservations(user) # is this bad?
```
judtinzhang, line 457: I think this is fine. Let's just keep in mind the `/recent` URL we have in `views.py` which also returns reservations and is used every time a user opens Penn Mobile. Luckily that route doesn't check for reservations made outside of our API so we don't incur this cost during those calls.

backend/gsr_booking/api_wrapper.py
```diff
@@ -532,38 +311,175 @@ def get_affiliation(self, email):
...
+            return ret
+    
+        booking_ids = set([booking['booking_id'] for booking in ret])
+        wharton_bookings = [booking for booking in wharton_bookings if booking['booking_id'] not in booking_ids]
+        if len(wharton_bookings) == 0:
+            return ret    
+
+        wharton_gsr_datas = {gsr.gid: GSRSerializer(gsr).data for gsr in GSR.objects.filter(kind=GSR.KIND_WHARTON)}        
+        for booking in wharton_bookings:
+            if booking['booking_id'] in booking_ids:
```
judtinzhang, line 471: I think this condition is always true looking at the filter you did on L465

judtinzhang (commented):
Looks great! Minor changes + 1 question about getting wharton reservations for Penn Labs (I think your query requires using the `is_wharton` field

backend/gsr_booking/views.py
tuneerroy: hm not sure if this will ever be `None` since think the default is `AnonymousUser`. any reason we don't wanna just make the route Authenticated?

backend/gsr_booking/views.py
```diff
@@ -211,11 +209,12 @@ def get(self, request, lid, gid):
...
-            if request.user in group.members.all():
-                return Response(GB.get_availability(lid, gid, start, end, request.user, group))
-            else:
-                return Response(BW.get_availability(lid, gid, start, end, request.user))
+            # group = Group.objects.get(name="Penn Labs")
+            # if request.user in group.members.all():
+            #     return Response(GB.get_availability(lid, gid, start, end, request.user, group))
+            # else:
+            #     return Response(BW.get_availability(lid, gid, start, end, request.user))
+            return Response(GSRBooker.get_availability(lid, gid, start, end, request.user, request.user.booking_groups.filter(name="Penn Labs").first()))
```
tuneerroy, line 217: can we break this code up a little bit instead of trying to 1-liner it lol

backend/gsr_booking/views.py
```diff
@@ -233,11 +232,7 @@ def post(self, request):
         room_name = request.data["room_name"]
 
         try:
-            group = Group.objects.get(name="Penn Labs")
-            if request.user in group.members.all():
-                GB.book_room(gid, room_id, room_name, start, end, request.user, group)
-            else:
-                BW.book_room(gid, room_id, room_name, start, end, request.user)
+            GSRBooker.book_room(gid, room_id, room_name, start, end, request.user, request.user.booking_groups.filter(name="Penn Labs").first())
```
tuneerroy, line 235: same as above

backend/gsr_booking/views.py
```diff
@@ -268,15 +263,5 @@ class ReservationsView(APIView):
...
-        return Response(BW.get_reservations(request.user))
-
-
-class CreditsView(APIView):
-    """
-    Gets credits for a User
-    """
-
-    permission_classes = [IsAuthenticated]
+        return Response(GSRBooker.get_reservations(request.user, request.user.booking_groups.filter(name="Penn Labs").first()))
```
tuneerroy, line 266: same as above

backend/gsr_booking/api_wrapper.py
```diff
@@ -31,175 +37,66 @@ class CreditType(Enum):
...
-                    del availability["to"]
-
-            context = {"name": lc_gsr["name"], "gid": lc_gsr["cid"]}
-            context["rooms"] = [
-                {"room_name": x["name"], "id": x["id"], "availability": x["availability"]}
-                for x in lc_gsr["rooms"]
-            ]
-            return context
-
+class AbstractBookingWrapper(ABC):
```
tuneerroy, line 40: holy cow, i haven't seen a Python abstract class in a while LOL

tuneerroy: nit: for all the methods, raise a `NotImplemented` exception instead of `pass`, unless this is just like a partial interface?

backend/gsr_booking/api_wrapper.py
```diff
@@ -31,175 +37,66 @@ class CreditType(Enum):
...
+    def book_room(self, rid, start, end, user):
+        """Books room if pennkey is valid"""
+        payload = {
+            "start": start,
+            "end": end,
+            "pennkey": user.username,
+            "room": rid,
+        }
+        url = f"{WHARTON_URL}{user.username}/student_reserve"
+        response = self.request("POST", url, json=payload).json()
```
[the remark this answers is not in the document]
tuneerroy: could be a util helper function

backend/gsr_booking/api_wrapper.py
tuneerroy: this works. Python momes haha

backend/gsr_booking/api_wrapper.py
tuneerroy: huh gonna have to explain this query to me one day.

backend/gsr_booking/api_wrapper.py
tuneerroy: better yet, let's just make a separate python file that returns these checks, just abstract our rules away so they are easy to change later

backend/gsr_booking/api_wrapper.py
```diff
@@ -532,38 +311,175 @@ def get_affiliation(self, email):
...
+
+    def book_room(self, gid, rid, room_name, start, end, user, group=None):
+        # NOTE when booking with a group, we are only querying our db for existing bookings, so users in a group who book through wharton may screw up the query
+        gsr = get_object_or_404(GSR, gid=gid)
+        start=datetime.datetime.strptime(start, "%Y-%m-%dT%H:%M:%S%z")
```
tuneerroy, line 358: ideally extract these `strptime` stuff to helper methods

backend/gsr_booking/api_wrapper.py
```diff
@@ -532,38 +311,175 @@ def get_affiliation(self, email):
...
+                if (curr_start := curr_end) >= end:
+                    break
+        except APIError as e:
+            raise APIError(f"{str(e)}. Was only able to book {start.strftime('%H:%M')} - {curr_start.strftime('%H:%M')}")
+
+        return reservation
+    
+    def cancel_room(self, booking_id, user):
+        # I think this should prefetch reservation, reservation__gsrbooking_set, and gsr ?
+        if gsr_booking := GSRBooking.objects.filter(booking_id=booking_id).prefetch_related(Prefetch('reservation__gsrbooking_set'), Prefetch('gsr')).first():
```
tuneerroy, line 402: nice! :D

backend/gsr_booking/api_wrapper.py
tuneerroy: lets gooooo HOF in the house

tuneerroy (commented):
Lot of code haha. Will do a round 2 in the future. Looks a lot cleaner than before tho, so nice!

backend/gsr_booking/api_wrapper.py
```diff
@@ -31,175 +37,66 @@ class CreditType(Enum):
...
+    
         try:
-            response = requests.request(*args, **kwargs)
+            response = requests.request(*arg, **kwargs)
         except (ConnectTimeout, ReadTimeout, ConnectionError):
             raise APIError("Wharton: Connection timeout")
 
         # only wharton students can access these routes
         if response.status_code == 403 or response.status_code == 401:
             raise APIError("Wharton: GSR view restricted to Wharton Pennkeys")
```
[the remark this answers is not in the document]
vcai122: should we just do != 200 raise APIError instead as well then? instead of just catching 403 and 401?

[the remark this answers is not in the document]
tuneerroy: that makes sense to me. just make sure there's also some message noting that it is a GSR request error tho.

backend/gsr_booking/api_wrapper.py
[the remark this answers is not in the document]
vcai122: did u ever push the util file u made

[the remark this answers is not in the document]
tuneerroy: yeah:  https://github.com/pennlabs/penn-mobile/tree/master/backend/utils

backend/gsr_booking/views.py
judtinzhang: Oh I was thinking we shouldn't make this Authenticated because I remember on iOS you could still view GSRs even when not logged in? But thinking about it some more I think making this authenticated is a good idea

backend/gsr_booking/api_wrapper.py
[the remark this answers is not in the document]
judtinzhang: agree, and also put the response's error message too so we know where it went wrong

backend/gsr_booking/api_wrapper.py
vcai122: smart smart

backend/gsr_booking/api_wrapper.py
vcai122: think copilot suggested

backend/gsr_booking/api_wrapper.py
vcai122: no need anymore if we check ok in the request first right

backend/gsr_booking/api_wrapper.py
vcai122: idk this was based off of the logic that was here previously

backend/gsr_booking/api_wrapper.py
vcai122: we are still using is_wharton for groups, which we would manage ourselves since we only have the 1 Penn Labs group. 
Otherwise our queries would be really expensive.

backend/gsr_booking/api_wrapper.py
vcai122: so 6?

backend/gsr_booking/api_wrapper.py
vcai122: tuneer says its the best way

backend/gsr_booking/api_wrapper.py
```diff
@@ -532,38 +309,239 @@ def get_affiliation(self, email):
...
+        return ret
+
+    def check_credits(self, user):
+        pass  # seems like its unused on the frontend
+
+
+# initialize singletons
+WhartonGSRBooker = WhartonBookingWrapper()
+LibCalGSRBooker = LibCalBookingWrapper()
+GSRBooker = BookingHandler(WhartonGSRBooker, LibCalGSRBooker)
```
judtinzhang, line 547: can we do something like this https://www.geeksforgeeks.org/singleton-pattern-in-python-a-complete-guide/#

vcai122: isn't this the Module-level Singleton described there? Are you saying a different type of singleton? I think the module level is the easiest/cleanest for our purposes

judtinzhang (approved):
Assuming only changes were just tests since the last review, LGTM! Good with merging in spite of test cases

PR #227, opened 2023-11-09
Temporary Laundry Patch 

tuneerroy:
Previous attempt at a solution failed. This is a dumber but more guaranteed way to avoid the error. Will investigate and revert later once root issue is identified.


tuneerroy:
> Seems reasonable to me haha. Can we confirm with someone on iOS that this increase in data size doesn't affect their code before pushing?

Jordan says it's fine. If it's breaks: his fault.

PR #231, opened 2023-12-03
GSR Location Caching

ashleyzhang01:


backend/gsr_booking/urls.py
```diff
@@ -23,7 +25,8 @@
 
 urlpatterns = [
     path("", include(router.urls)),
-    path("locations/", Locations.as_view(), name="locations"),
+    path("locations/", Locations.as_view(),
```
judtinzhang, line 28: Oh that is smart, much better than implementing the `list` function, thanks!

judtinzhang (commented):
Might need to fix some merging, but functionally LGTM, thanks!

PR #234, opened 2023-12-08
Add Sublet Images

dr-Jess:
PR to implement image handling functionality to new Sublet app.
- [x] Upload images when creating sublet
- [x] Get image URLS only when retrieving/listing
- [x] Add images when editing sublet
- [x] Delete images by id when editing sublet
- [x] Test cases


backend/sublet/models.py
```diff
@@ -42,6 +42,7 @@ class Sublet(models.Model):
     baths = models.IntegerField(null=True, blank=True)
     description = models.TextField(null=True, blank=True)
     external_link = models.URLField(max_length=255)
+    # TODO: change to just one price field and migrateeeee
```
judtinzhang, line 45: Why do we want to range price field? I recall wanting 2 to give a range since the sublet price can be negotiable

backend/sublet/serializers.py
```diff
@@ -51,54 +49,96 @@ def get_image_url(self, obj):
...
     class Meta:
         model = Sublet
-        exclude = ["favorites"]
-        read_only_fields = ["id", "created_at", "subletter", "sublettees"]
+        read_only_fields = [
+            "id",
+            "created_at",
+            "subletter",
+            "sublettees",
+            # "images"
```
judtinzhang, line 71: nit: in addition to leaving this here, could you also leave a brief comment saying how `images` are created separately + are shown in the Read serializer?

backend/sublet/serializers.py
```diff
@@ -51,54 +49,96 @@ def get_image_url(self, obj):
...
 
-    def parse_amenities(self, raw_amenities):
-        if isinstance(raw_amenities, list):
-            ids = raw_amenities
-        else:
-            ids = (
-                list() if len(raw_amenities) == 0 else [str(id) for id in raw_amenities.split(",")]
-            )
-        return Amenity.objects.filter(name__in=ids)
+    # def parse_amenities(self, raw_amenities):
```
judtinzhang, line 91: I'm down to remove this code entirely if the tests pass

backend/sublet/serializers.py
```diff
@@ -51,54 +49,96 @@ def get_image_url(self, obj):
...
+    #         ids = (
+    #         list() if len(raw_amenities) == 0 else [str(id) for id in raw_amenities.split(",")]
+    #         )
+    #     return Amenity.objects.filter(name__in=ids)
 
     def create(self, validated_data):
         validated_data["subletter"] = self.context["request"].user
+        # images = validated_data.pop("images") if "images" in validated_data else []
         instance = super().create(validated_data)
-        data = self.context["request"].POST
```
judtinzhang, line 78: ditto here, down to remove the commented out code

backend/sublet/serializers.py
```diff
@@ -51,54 +49,96 @@ def get_image_url(self, obj):
...
         if (
             self.context["request"].user == instance.subletter
             or self.context["request"].user.is_superuser
         ):
-            amenities_data = self.context["request"].data
-            if amenities_data.get("amenities") is not None:
-                amenities = self.parse_amenities(amenities_data.getlist("amenities"))
-                instance.amenities.set(amenities)
-            validated_data.pop("amenities", None)
+            # amenities_data = self.context["request"].data
```
judtinzhang, line 124: ditto here

backend/sublet/serializers.py
```diff
@@ -110,8 +150,34 @@ def destroy(self, instance):
...
+            "start_date",
+            "end_date",
+            "expires_at",
+            "images",
+        ]
+
+
 # simple sublet serializer for use when pulling all serializers/etc
-class SimpleSubletSerializer(serializers.ModelSerializer):
+class SubletSerializerSimple(serializers.ModelSerializer):
```
judtinzhang, line 181: Do we still need this serializer now that we have `SubletSerializerRead`?  Seems like the only difference is the read only fields, but I don't think that matters after looking at the simpler serializer's use cases.

backend/sublet/views.py
```diff
@@ -62,11 +71,37 @@ class Properties(viewsets.ModelViewSet):
...
 
+    def create(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)  # Check if the data is valid
+        instance = serializer.save()  # Create the Sublet
+        instance_serializer = SubletSerializerRead(instance=instance, context={"request": request})
+        return Response(instance_serializer.data, status=status.HTTP_201_CREATED)
+
+    def update(self, request, *args, **kwargs):
+        partial = kwargs.pop("partial", False)
```
judtinzhang, line 90: What is `partial` here? Is this some internal django field?

backend/sublet/views.py
```diff
@@ -104,14 +139,14 @@ def list(self, request, *args, **kwargs):
...
 
         queryset = Sublet.objects.all().filter(expires_at__gte=timezone.now())
-
         # Apply filters based on query parameters
         if title:
             queryset = queryset.filter(title__icontains=title)
         if address:
             queryset = queryset.filter(address__icontains=address)
         if amenities:
-            queryset = queryset.filter(amenities__name__in=amenities)
```
judtinzhang, line 114: I believe the deleted line is the functionality we want. If we provide a `amenities=["X", "Y"]` then the for loop you implemented would filter for `amenities__name="X" and amenities__name="Y"`, which will lead to the empty set.

judtinzhang (commented):
Almost there! Few nits / questions

judtinzhang:
One last comment -- could we also test deleting an image / making sure the other images are still preserved

backend/sublet/models.py
dr-Jess: I think vincent/ashley/I discussed with the other teams and the consensus was the prices were negotiable anyway, so min/max doesn't really serve a purpose. Def open to reconsideration for either way though.

backend/sublet/serializers.py
dr-Jess: Good point; the original idea of the simple version is to send only the data necessary for a browse view (as opposed to viewing a singular sublet), but not sure if the few text fields being saved are worth. Thoughts @vcai122?

backend/sublet/views.py
dr-Jess: I think it's used by the serializer to differentiate patch and put? Just a guess but that makes the most sense.

backend/sublet/views.py
dr-Jess: Uhhhh not sure, I got this from @vcai122 thoughts?

dr-Jess: Gut feeling is that you're right Justin, but maybe there's a reason Vincent asked me to make this change.

backend/sublet/serializers.py
```diff
@@ -51,54 +50,66 @@ def get_image_url(self, obj):
...
     class Meta:
         model = SubletImage
-        fields = ["image_url"]
+        fields = ["id", "image_url"]
 
 
 # complex sublet serializer for use in C/U/D + getting info about a singular sublet
 class SubletSerializer(serializers.ModelSerializer):
-    amenities = AmenitySerializer(many=True, required=False)
+    # amenities = AmenitySerializer(many=True, required=False)
```
judtinzhang, line 58: nit: remove comments

backend/sublet/views.py
judtinzhang: ah, makes sense

judtinzhang (commented):
LGTM, thanks jesse! let's wait for vincent's sign off + working out the price before push

backend/sublet/serializers.py
vcai122: discussed - worth leaving since we already made it

backend/sublet/views.py
```diff
@@ -130,12 +166,55 @@ def list(self, request, *args, **kwargs):
...
+
+
+class DeleteImage(generics.DestroyAPIView):
+    serializer_class = SubletImageSerializer
+    http_method_names = ["delete"]
+    permission_classes = [SubletImageOwnerPermission | IsSuperUser]
+    queryset = SubletImage.objects.all()
+
+    def destroy(self, request, *args, **kwargs):
+        queryset = self.get_queryset()
```
vcai122, line 209: `nit`: no need to extract to queryset variable since only used once and relatively simple lol

backend/sublet/views.py
vcai122: justin said its fine

vcai122 (changes requested):
LGTM

vcai122 (approved):

PR #241, opened 2024-02-04
Calendar Performance Update

judtinzhang:


backend/penndata/management/commands/get_calendar.py
```diff
@@ -0,0 +1,77 @@
...
+                            last_date = date_info.split("-")[0].split(" ")
+                            month = last_date[0]
+                            day = last_date[1]
+                            date = datetime.datetime.strptime(
+                                month + day + str(current_year) + "-04:00", "%B%d%Y%z"
+                            )
+                        except (ValueError, IndexError):
+                            pass
+
+                if date and current_time.replace(tzinfo=None) < date.replace(tzinfo=None) < (
```
judtinzhang, line 69: @ashleyzhang01 could you explain this line here?

backend/penndata/models.py
```diff
@@ -58,3 +58,14 @@ class AnalyticsEvent(models.Model):
...
     def __str__(self):
         return f"{self.cell_type}-{self.user.username}"
+
+
+class CalendarEvent(models.Model):
+    event = models.CharField(max_length=255)
+    date = models.CharField(max_length=50, null=True, blank=True)
+    # NOTE: This is bad practice, though is necessary for the time being
+    # since frontends use the string date field
+    date_obj = models.DateTimeField(null=True, blank=True)
```
judtinzhang, line 68: We probably need to change this in the future.

backend/penndata/views.py
```diff
@@ -70,80 +79,19 @@ def get(self, request):
...
-        return calendar
+    permission_classes = [AllowAny]
+    serializer_class = CalendarEventSerializer
 
-    def get(self, request):
-        return Response(self.get_calendar())
+    def get_queryset(self):
+        return CalendarEvent.objects.filter(
+            date_obj__gte=timezone.localtime(),
+            date_obj__lte=timezone.localtime() + timedelta(days=30),
```
judtinzhang, line 93: For now we need this because we never clear the database. @ashleyzhang01 what do you think? Do you think the `manage.py` command should delete all previous `CalendarEvents` that way we don't have to re-do this filtering for events?

backend/penndata/management/commands/get_calendar.py
ashleyzhang01: I set the time zone for current_time to none because couldn't compare the datetime.datetime.now() with the date due to some time zone error. I guess if we moved back to the Django timezone, we might not have to do this since when we were using it before, that wasn't an issue?

backend/penndata/models.py
ashleyzhang01: Should have fields for a start date and a nullable end date bc some calendar events are ranges.

backend/penndata/views.py
ashleyzhang01: We don't call the manage.py thing often enough that it deletes all previous. I had something in my Calendar method in the views.py for this-- can you move it in too? 
Basically, it deletes any previous events when called each time, so we don't have to do the manage.py as often and don't have to filter through the time range as often.

ashleyzhang01, line 93: Wait also if it is only called once a month, then by the end of the month there won't be many future events left in the db, so we wouldn't really have any calendar events?

tuneerroy (commented):
looking good :)

backend/penndata/management/commands/get_calendar.py
judtinzhang: Sounds good, yea will revert back to timezone

backend/penndata/models.py
judtinzhang: Thanks for the reminder. I will leave this for you to do in a future events PR you create so we can group relevant code

backend/penndata/views.py
judtinzhang: thanks, just did

judtinzhang: Ah yes good point. I'll change the `get_calendar` to create `CalendarEvents` that exceed 30 days as well. Then we will leave this view as is.

backend/penndata/management/commands/get_calendar.py
```diff
@@ -0,0 +1,77 @@
...
+from django.core.management.base import BaseCommand
+
+from penndata.models import CalendarEvent
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **kwargs):
+        # scrapes almanac from upenn website
+        try:
+            resp = requests.get("https://almanac.upenn.edu/penn-academic-calendar")
```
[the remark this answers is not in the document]
judtinzhang: sounds gooood tanner :D

backend/penndata/management/commands/get_calendar.py
```diff
@@ -0,0 +1,77 @@
...
+
+import requests
+from bs4 import BeautifulSoup
+from django.core.management.base import BaseCommand
+
+from penndata.models import CalendarEvent
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **kwargs):
```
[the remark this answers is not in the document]
judtinzhang: agreed. just cleaned it up to make it more readable.

ashleyzhang01 (approved):
lgtm

PR #242, opened 2024-02-15
Fix wharton availability timezone issue

vcai122:


backend/gsr_booking/api_wrapper.py
```diff
@@ -116,13 +116,14 @@ def get_availability(self, lid, start, end, user):
 
         # presets end date as end midnight of next day
         end_date = (
-            datetime.datetime.strptime(end + "T00:00:00-04:00", "%Y-%m-%dT%H:%M:%S%z")
```
judtinzhang, line 119: Quick q: should we also add a  `timedelta(days=1)` here, otherwise this `end_date` isn't going to be inclusive of that day

judtinzhang (approved):

backend/gsr_booking/api_wrapper.py
vcai122: hm yea make sense so if they pass in end_date=start_date we give them just that info for that day?
although seems like we weren't doing this before

judtinzhang: Yeaa can you quickly scan the ios and android apps do see if they use it? if they don't then let's change it to be inclusive to make this future proof lol

PR #244, opened 2024-02-23
Events scraping

ashleyzhang01:
guess how many shots of vodka


backend/penndata/management/commands/get_events.py
```diff
@@ -0,0 +1,104 @@
...
+                    start_date = start_date.replace(year=current_year + 1)
+            if start_time_str == "All day":
+                start_time = datetime.time(0, 0)
+            else:
+                start_time = datetime.datetime.strptime(start_time_str, "%I:%M%p").time()
+            start_date = datetime.datetime.combine(start_date, start_time)
+
+            event_url = urljoin(PENN_TODAY_WEBSITE, article.find("a", class_="tease__link")["href"])
+
+            Event.objects.update_or_create(
```
dr-Jess, line 89: Given that we've previously wiped all events, is there a reason to use `update_or_create` here?

backend/penndata/models.py
```diff
@@ -9,15 +9,16 @@
 
 
 class Event(models.Model):
-    event_type = models.CharField(max_length=255)
+    event_type = models.CharField(max_length=255, null=True, blank=True)
```
dr-Jess, line 12: What kinds of types are we storing here?

dr-Jess (commented):
Good stuff 👍 but def run by Justin :P

judtinzhang:
9

backend/penndata/views.py
```diff
@@ -104,7 +104,16 @@ class Events(generics.ListAPIView):
...
     def get_queryset(self):
-        return Event.objects.filter(event_type=self.kwargs.get("type", ""))
+        queryset = Event.objects.all()
+
+        event_type = self.kwargs.get("type")
+        if event_type:
+            queryset = queryset.filter(event_type=event_type)
+
+        queryset = queryset.filter(
+            end__gte=timezone.localtime(), start__lte=timezone.localtime() + timedelta(days=30)
```
judtinzhang, line 114: why does `start` have to be less than 30 days from now?

backend/penndata/models.py
```diff
@@ -9,15 +9,16 @@
...
-    website = models.URLField(max_length=255, null=True)
-    facebook = models.URLField(max_length=255, null=True)
+    description = models.TextField(null=True, blank=True)
+    image_url = models.URLField(null=True, blank=True)
+    start = models.DateTimeField()
+    end = models.DateTimeField(null=True, blank=True)
+    location = models.CharField(max_length=255, null=True, blank=True)
+    email = models.CharField(max_length=255, null=True, blank=True)
+    website = models.URLField(max_length=255, null=True, blank=True)
+    facebook = models.URLField(max_length=255, null=True, blank=True)
```
judtinzhang, line 21: Tbh I would be down to remove the `facebook` field. seems too specific

backend/penndata/management/commands/get_events.py
```diff
@@ -0,0 +1,104 @@
...
+
+from penndata.models import Event
+
+
+PENN_TODAY_WEBSITE = "https://penntoday.upenn.edu/events"
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **kwargs):
+        now = datetime.datetime.now()
```
judtinzhang, line 19: nit: Can we change this to django `timezone.localtime()` so we are more precise with our datetimes (local vs utc)

backend/penndata/management/commands/get_events.py
```diff
@@ -0,0 +1,104 @@
...
+                start_time_str = "All day"
+                location = None
+
+            end_date_elem = article.find(
+                "p", class_="tease__meta--sm", string=lambda x: "Through" in str(x)
+            )
+            end_date = None
+            if end_date_elem:
+                end_date_str = end_date_elem.text.strip().split(" ")[-1]
+                end_date = (
```
judtinzhang, line 63: nit: not sure if you like this better, but what do you think of:
```python3
from datetime import time
end_of_day = time(23, 59, 59)
end_date = datetime.datetime.strptime(end_date_str, "%m/%d/%Y")
end_datetime = datetime.combine(end_date, end_of_day)
```

backend/penndata/management/commands/get_events.py
```diff
@@ -0,0 +1,104 @@
...
+                    year=current_year
+                ) + datetime.timedelta(days=1)
+            else:
+                start_date = datetime.datetime.strptime(start_date_str, "%m/%d").replace(
+                    year=current_year
+                )
+                if start_date.month < current_month:
+                    # If scraped month is before current month, increment year
+                    start_date = start_date.replace(year=current_year + 1)
+            if start_time_str == "All day":
```
judtinzhang, line 81: Can we pull out this `"All Day"` into some sort of global variable, just so we don't need to hard code magic strings into the code?

backend/penndata/management/commands/get_events.py
```diff
@@ -0,0 +1,104 @@
...
+            end_date = None
+            if end_date_elem:
+                end_date_str = end_date_elem.text.strip().split(" ")[-1]
+                end_date = (
+                    datetime.datetime.strptime(end_date_str, "%m/%d/%Y")
+                    + datetime.timedelta(days=1)
+                    - datetime.timedelta(seconds=1)
+                )
+
+            if start_date_str == "02/29" or start_date_str == "2/29":
```
judtinzhang, line 69: Any reason we needed to split the logic up if it is leap year?

judtinzhang (commented):
Good stuff! Besides code comments, I would recommend renaming the `get_events` command to `get_penn_today_events` just cause this script is specific to Penn Today.

backend/penndata/views.py
ashleyzhang01: events displayed should start within next 30 days, if they haven't started yet. if they have started already, then only show if they haven't ended yet

backend/penndata/management/commands/get_events.py
ashleyzhang01: some issue happens with the year because we handle that separately, since the date doesn't automatically include year. So, if it's a leap year, it won't process until after the string date is turned into month day, which will make it break

backend/penndata/management/commands/get_penn_today_events.py
```diff
@@ -0,0 +1,142 @@
...
+        for article in event_articles:
+            name = article.find("h3", class_="tease__head").text.strip()
+            description = article.find("div", class_="tease__dek").text.strip()
+            start_date_str = article.find("p", class_="tease__date").text.strip()
+            meta_elements = article.find_all("p", class_="tease__meta--sm")
+            if len(meta_elements) >= 2:
+                start_time_str = meta_elements[0].text.strip().replace(".", "")
+                location = meta_elements[1].text.strip()
+            else:
+                start_time_str = "all day"
```
vcai122, line 56: should we use ALL_DAY constant here?

backend/penndata/management/commands/get_penn_today_events.py
```diff
@@ -0,0 +1,142 @@
...
+                start_time = datetime.time(0, 0)
+            else:
+                start_time = datetime.datetime.strptime(start_time_str, "%I:%M%p").time()
+            start_date = datetime.datetime.combine(start_date, start_time)
+
+            event_url = urljoin(PENN_TODAY_WEBSITE, article.find("a", class_="tease__link")["href"])
+
+            end_time = self.get_end_time(event_url)
+            if end_time is not None:
+                if end_date_elem:  # end date and end time
```
vcai122, line 87: small nit (ignore, I'm mostly just curious about @judtinzhang thoughts)

I have a strong preference to put this all in 1 line since you are just making 1 final target end_date and everything else is a temp variable.

vcai122 (approved):

PR #246, opened 2024-02-27
Priority field to Post/Polls + Refactor

vcai122:


backend/portal/models.py
```diff
@@ -37,15 +37,22 @@ class Poll(models.Model):
     )
 
     club_code = models.CharField(max_length=255, blank=True)
```
judtinzhang, line 39: Hmm, not entirely sure of this myself. What are people's thoughts on having `club_code` and `club_comment` here? Do you think this is general enough for future models or should we push this back down to the `Poll` and `Post` models.

backend/portal/models.py
```diff
@@ -37,15 +37,22 @@ class Poll(models.Model):
...
+    expire_date = models.DateTimeField()
     status = models.CharField(max_length=30, choices=STATUS_OPTIONS, default=STATUS_DRAFT)
-    multiselect = models.BooleanField(default=False)
     club_comment = models.CharField(max_length=255, null=True, blank=True)
     admin_comment = models.CharField(max_length=255, null=True, blank=True)
     target_populations = models.ManyToManyField(TargetPopulation, blank=True)
+    priority = models.IntegerField(default=0)
+
+    class Meta:
+        abstract = True
```
judtinzhang, line 50: Good stuff haha

backend/portal/views.py
```diff
@@ -99,7 +99,7 @@ def get_queryset(self):
             )
         )
 
-    @action(detail=False, methods=["post"])
+    @action(detail=False, methods=["post"])  # WHY IS THIS POST
```
judtinzhang, line 102: `POST` because the devices send us a `id_hash` which is a hash based on device id. This is because we want to associate votes based on this rather than pennkey to keep users anonymous. For this, if the user voted we don't given them the ability to vote again (based on if `id_hash` already voted)

judtinzhang (commented):
LGTM! Just one comment worth considering

backend/portal/models.py
vcai122: wait what future models were you thinking of? I was thinking Content is just the "superclass" to Post/Polls, so has all the shared stuff. I guess you're thinking of a different "meaning" to Content?

backend/portal/views.py
vcai122: wait wat, but i don't see any posting going on here. am i blind? it looks like its just updating stuff

backend/portal/models.py
judtinzhang: Hm that is fair. Yea I think posts and polls are really all that is it. I was thinking maybe Ashley's Events could also inherit but thinking about it some more that wouldn't make sense. Okay down to merge haha

backend/portal/views.py
judtinzhang: L108 ` id_hash = request.data["id_hash"]`

vcai122: but there's no up updating to our models going on? like can't we just use a get request for this

vcai122: oh wait nvm. https://stackoverflow.com/questions/19637459/rest-api-using-post-instead-of-get. so we can send through the body im assuming?

judtinzhang: Yea right. I think generally I like sending things via POST body as opposed to query params if it isn't literally for a non trivial query

PR #247, opened 2024-02-28
Fix libcal booking

vcai122:


judtinzhang (approved):
LOL good stuff thanks vincent!

backend/gsr_booking/models.py
```diff
@@ -96,8 +101,13 @@ class GSR(models.Model):
     name = models.CharField(max_length=255)
     image_url = models.URLField()
 
+    in_use = models.BooleanField(default=True)
+
+    objects = GSRManager()
+    all_objects = models.Manager()  # for admin page
```
judtinzhang, line 107: that is great, thanks vincent! @tuneerroy @ashleyzhang01 @JeffersonDing @dr-Jess just for reference look at this man!

backend/tests/gsr_booking/test_gsr_views.py
```diff
@@ -82,8 +82,8 @@ def test_recent(self):
         response = self.client.get(reverse("recent-gsrs"))
         res_json = json.loads(response.content)
         self.assertEquals(2, len(res_json))
-        self.assertEquals(6, len(res_json[0]))
-        self.assertEquals(6, len(res_json[1]))
+        self.assertEquals(7, len(res_json[0]))
+        self.assertEquals(7, len(res_json[1]))
```
judtinzhang, line 86: i'm imagining that the extra field is the `in_use`? any reason it still shows up in the route despite all of our queries using `.object` to query? if this is the case, I think we should remove this field from showing up

judtinzhang (commented):
One quick question before push

backend/tests/gsr_booking/test_gsr_views.py
vcai122: oh the serializer was still serializing all the fields. fixed now

judtinzhang (approved):

backend/tests/gsr_booking/test_gsr_views.py
judtinzhang: bet LGTM

PR #251, opened 2024-03-05
Change Sublet Amenity Return Formatting

dr-Jess:
Previously, amenities were returned on Sublets as follows:
```
"amenities":[
    {
    "name":"amenityname"
    },
    {
    "name":"amenityname2"
    }
]
```
This pull request is to modify to a new formatting as follows:
```
"amenities":["amenityname","amenityname2"]
```
Additionally, this pull request migrates the "baths" field on sublets to allow more one-decimal place numbers.


backend/sublet/serializers.py
```diff
@@ -122,7 +122,9 @@ def destroy(self, instance):
 
 
 class SubletSerializerRead(serializers.ModelSerializer):
-    amenities = AmenitySerializer(many=True, required=False)
+    amenities = serializers.PrimaryKeyRelatedField(
```
judtinzhang, line 125: Thanks! Can we confirm this change by adding an additional check in one of our test cases? Should just be one additional `assert` or something to assert that the type now returned is a list of strings

judtinzhang (commented):
Code looks good! One comment about tests and also current tests are failing (looks like its from decimal change). Suggest running tests locally once before pushing in the future

backend/sublet/serializers.py
dr-Jess: Fixed! https://github.com/pennlabs/penn-mobile/pull/251/commits/19a351c49fd87dedcaf5fecc4aa058dcbf72d7f8 checks this by a direct comparison to the expected string list output.

judtinzhang (approved):

PR #252, opened 2024-03-06
More events scraping

ashleyzhang01:
adding calendar events from across Penn (venture lab, engineering, college houses, wharton)


backend/penndata/management/commands/get_engineering_events.py
```diff
@@ -0,0 +1,67 @@
...
+                    "start": timezone.make_aware(start),
+                    "end": timezone.make_aware(end),
+                    "location": location,
+                    "website": url,
+                    "description": description,
+                    "email": email,
+                },
+            )
+
+        self.stdout.write("Uploaded Engineering Events!")
```
judtinzhang, line 67: This looks great, thanks Ashley! One thing I'm curious about is why you chose the `html` library instead of `BeautifulSoup`? Wondering if the library could make the parsing easier/cleaner, but good with either!

backend/penndata/management/commands/get_rodin_events.py
```diff
@@ -0,0 +1,87 @@
...
+from bs4 import BeautifulSoup
+from django.core.management.base import BaseCommand
+from django.utils import timezone
+
+from penndata.models import Event
+
+
+RODIN_EVENTS_WEBSITE = "https://rodin.house.upenn.edu"
+HEADERS = {
+    "User-Agent": "Mozilla/5.0 AppleWebKit/537.36 Chrome/91.0.4472.124 Safari/537.36"
```
judtinzhang, line 13: Woah any reason for this? haha

backend/penndata/management/commands/get_rodin_events.py
```diff
@@ -0,0 +1,87 @@
...
+}
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **kwargs):
+        self.scrape_calendar_page(RODIN_EVENTS_WEBSITE + "/calendar")
```
judtinzhang, line 19: suggest using python's `f-string` here instead of the `+` operator

backend/penndata/management/commands/get_rodin_events.py
```diff
@@ -0,0 +1,87 @@
...
+
+        event_cells = soup.find_all("td", class_="single-day future")
+
+        for cell in event_cells:
+            item = cell.find("div", class_="item")
+            if item:
+                event_link = item.find("a", href=True)
+                if event_link:
+                    name = event_link.text.strip()
+                    url = RODIN_EVENTS_WEBSITE + event_link["href"]
```
judtinzhang, line 66: nit: suggest `f-string`

backend/penndata/management/commands/get_rodin_events.py
```diff
@@ -0,0 +1,87 @@
...
+            resp = requests.get(calendar_url, headers=HEADERS)
+        except ConnectionError:
+            return None
+        soup = BeautifulSoup(resp.text, "html.parser")
+
+        event_cells = soup.find_all("td", class_="single-day future")
+
+        for cell in event_cells:
+            item = cell.find("div", class_="item")
+            if item:
```
judtinzhang, line 62: nit: personally I'm not a big fan of a lot of nesting because it ends up being hard to read the more layers you get. I would suggest doing something like:
```python3
if not item:
   continue
event_link = item.find("a", href=True)
if not event_link:
   continue
```
That way we can avoid nesting

backend/penndata/management/commands/get_engineering_events.py
```diff
@@ -0,0 +1,67 @@
...
+                defaults={
+                    "event_type": "Engineering",
+                    "image_url": "",
+                    "start": timezone.make_aware(start),
+                    "end": timezone.make_aware(end),
+                    "location": location,
+                    "website": url,
+                    "description": description,
+                    "email": email,
+                },
```
judtinzhang, line 64: Do we need some sort of check to only add events within 30 days from now?

backend/penndata/management/commands/get_rodin_events.py
```diff
@@ -0,0 +1,87 @@
...
+    "User-Agent": "Mozilla/5.0 AppleWebKit/537.36 Chrome/91.0.4472.124 Safari/537.36"
+}
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **kwargs):
+        self.scrape_calendar_page(RODIN_EVENTS_WEBSITE + "/calendar")
+        now = timezone.localtime()
+        current_day, current_month, current_year = now.day, now.month, now.year
+        if current_day > 25:
```
judtinzhang, line 22: nit: I would suggest utilizing the `datetime` library to calculate next month's year and stuff for you. I don't know the precise syntax, but I'd imagine you can do something like:
```python3
from datetime import timedelta
curr = timezone.localtime()
next = timezone.localtime() + timedelta(months=1)
next_day, next_month, next_year = next.day, next.month, next.year
```
I guess don't quote me on this, I would suggest checking what this kind of code does for like months that don't have a 31'st, but I feel like this might be better in terms of readability / scalability

judtinzhang (commented):
Guess: 0 shots.

Looks good! I didn't check `get_venture_events` because I didn't want to completely bombard this PR. But when you get a chance to look at the comments I made and make any revisions, feel free to ping me again!

backend/penndata/management/commands/get_engineering_events.py
ashleyzhang01: So for the engineering events, they are all in a json format under <script type="application/ld+json">, so I use the html library to replace characters (such as &lt;p&gt; for "). The rest is Python to get the json content I'm pretty sure

backend/penndata/management/commands/get_rodin_events.py
ashleyzhang01: some sites block scraping unless you use headers. they're needed for the venture lab one i think, but i put them in for the college houses just in case. should i only keep in the venture lab one?

backend/penndata/management/commands/get_engineering_events.py
ashleyzhang01: by default, they only show events for next week or so. i played around with it a bit and it gets janky for trying to get events up to a certain date so I thought it wasn't necessary to get events that far ahead

backend/penndata/management/commands/get_rodin_events.py
judtinzhang: sounds good, thanks! and yes, I think it would probably be best to just put it on the files that require it

backend/penndata/management/commands/get_rodin_events.py
vcai122: could also use the walrus
`if (item := cell.find("div", class_="item")) and (event_link := item.find("a", href=True))`

backend/penndata/management/commands/get_engineering_events.py
```diff
@@ -0,0 +1,65 @@
...
+        start_marker = '<script type="application/ld+json">'
+        end_marker = "</script>"
+        start_index = html_content.find(start_marker)
+        end_index = html_content.find(end_marker, start_index)
+        json_ld_content = html_content[start_index + len(start_marker) : end_index]
+
+        events_data = json.loads(json_ld_content)
+
+        for event in events_data:
+            event_name = html.unescape(event.get("name", ""))
```
judtinzhang, line 34: If `event_name` here is blank, do you think we should still create the `Event` object? I feel like the name is necessary

backend/penndata/management/commands/get_engineering_events.py
```diff
@@ -0,0 +1,65 @@
...
+            description = (
+                html.unescape(event.get("description", "")).replace("<p>", "").replace("</p>\n", "")
+            )
+            url = event.get("url", "")
```
judtinzhang, line 38: nit: suggest change `url = event.get("url", None)`. An empty string would not pass the checks for a `URLField`

backend/penndata/management/commands/get_engineering_events.py
```diff
@@ -0,0 +1,65 @@
...
+
+            email = None
+            if event.get("organizer"):
+                email = html.unescape(event.get("organizer").get("email"))
+
+            Event.objects.update_or_create(
+                name=event_name,
+                defaults={
+                    "event_type": Event.TYPE_PENN_ENGINEERING,
+                    "image_url": "",
```
judtinzhang, line 55: Same here, probably should make this `None`

backend/penndata/management/commands/get_engineering_events.py
```diff
@@ -0,0 +1,65 @@
...
+            description = (
+                html.unescape(event.get("description", "")).replace("<p>", "").replace("</p>\n", "")
+            )
+            url = event.get("url", "")
+
+            start = datetime.datetime.fromisoformat(event.get("startDate"))
+            end = datetime.datetime.fromisoformat(event["endDate"]) if "endDate" in event else None
+
+            location = None
+            if "location" in event:
```
judtinzhang, line 44: nit: suggest 
```python3
location = event.get("location", dict()).get("name")
```

backend/penndata/management/commands/get_engineering_events.py
```diff
@@ -0,0 +1,65 @@
...
+                location = event.get("location").get("name")
+
+            email = None
```
judtinzhang, line 47: nit: suggest changing this to something similar to the location code by passing in an empty dict as a default

backend/penndata/management/commands/get_college_house_events.py
```diff
@@ -0,0 +1,140 @@
...
+            if soup.select_one(".field-name-body")
+            else ""
+        )
+        return location, start_time, end_time, description
+
+    def scrape_calendar_page(self, calendar_url):
+        try:
+            resp = requests.get(calendar_url)
+        except ConnectionError:
+            return None
```
judtinzhang, line 103: nit: since this method inherently doesn't return anything, I would suggest just `return` instead of `return None`

backend/penndata/management/commands/get_college_house_events.py
```diff
@@ -0,0 +1,140 @@
...
+
+        for cell in event_cells:
+            item = cell.find("div", class_="item")
+            if not item:
+                continue
+            event_link = item.find("a", href=True)
+            if not event_link:
+                continue
+            name = event_link.text.strip()
+            url = event_link["href"]
```
judtinzhang, line 116: Can we be sure that `url` is here?

backend/penndata/management/commands/get_college_house_events.py
```diff
@@ -0,0 +1,140 @@
...
+            index = calendar_url.find("/", calendar_url.find("://") + 3)
+            base_url = calendar_url[:index]
+            url = f"{base_url}{url}"
+
+            location, start_time, end_time, description = self.scrape_details(url)
+
+            house = calendar_url.split("/")[2].split(".")[0]
+
+            event_type = EVENT_TYPE_MAP.get(house, None)
```
judtinzhang, line 125: Instead of building a list of websites and a dict from house to event type, can we have a dict from website to event type? That way, we can just directly look up the event type without extra logic

backend/penndata/management/commands/get_college_house_events.py
```diff
@@ -0,0 +1,140 @@
...
+                name=name,
+                defaults={
+                    "event_type": event_type,
+                    "image_url": "",
+                    "start": timezone.make_aware(start_time),
+                    "end": timezone.make_aware(end_time),
+                    "location": location,
+                    "website": url,
+                    "description": description,
+                    "email": "",
```
judtinzhang, line 136: For these and other `email` fields, I would suggest making them `None` rather than empty string just because `None` is the catch-all "there is nothing here". Since the DB fields are nullable we should be utilizing that

backend/penndata/management/commands/get_college_house_events.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def scrape_details(self, event_url):
+        try:
+            resp = requests.get(event_url)
+        except ConnectionError:
+            print("Error:", ConnectionError)
+            return None
+        soup = BeautifulSoup(resp.text, "html.parser")
+
+        location = (
```
judtinzhang, line 67: Same comments as before in terms of defaulting to `None`

backend/penndata/management/commands/get_venture_events.py
```diff
@@ -0,0 +1,112 @@
...
+            # some events don't have a start/end date time or a year
+            if event_date_elem:
+                event_date_str = event_date_elem.text.strip()
+
+                event_date_parts = event_date_str.split(" at ")
+                event_start_str = event_date_parts[1].split(" - ")[0].strip()
+                event_end_str = event_date_parts[1].split(" - ")[1].strip()
+
+                event_start_datetime = datetime.strptime(
+                    event_date_parts[0] + " " + event_start_str, "%B %d, %Y %I:%M%p"
```
judtinzhang, line 45: nit: suggest `f-string` here and below

backend/penndata/management/commands/get_venture_events.py
```diff
@@ -0,0 +1,112 @@
...
+                )
+                last_start_datetime = event_start_datetime
+            else:  # if no year given
+                event_month_elem = event.find("div", class_="PromoSearchResultEvent-month")
+                event_day_elem = event.find("div", class_="PromoSearchResultEvent-day")
+
+                if event_month_elem and event_day_elem:
+                    event_month = event_month_elem.text.strip()
+                    event_day = int(event_day_elem.text.strip())
+
```
judtinzhang, line 58: Makes a lot of sense, great work for solving out the logic + comments + making the code very readable!

backend/penndata/management/commands/get_college_house_events.py
```diff
@@ -0,0 +1,140 @@
...
+        now = timezone.localtime()
+        if now.day > 25:
+            next = now + datetime.timedelta(days=30)
+            next_month, next_year = next.month, next.year
+            for site in EVENT_SITES:
+                self.scrape_calendar_page(f"{site}/calendar/{next_year}-{next_month:02d}")
+
+        self.stdout.write("Uploaded Rodin College House Events!")
+
+    def scrape_details(self, event_url):
```
judtinzhang, line 59: Awesome, really like this helper it is very clean

backend/penndata/management/commands/get_venture_events.py
```diff
@@ -0,0 +1,112 @@
...
+                    )
+
+            # events are ordered from future to past, so break once we find a past event
+            if event_start_datetime < now:
+                break
+
+            print("Event Start:", event_start_datetime)
+            print("Event End:", event_end_datetime)
+
+            title = event.find("div", class_="PromoSearchResultEvent-title")
```
judtinzhang, line 82: I think here and the below lines might be a good use of a walrus operator to save a line per if-statement. See [here](https://realpython.com/python-walrus-operator/) for examples!

judtinzhang (commented):
Great work! Left a few comments, almost there! 

I left the following notes for posterity since we probably need to implement these eventually, but not now lol:
- Probably need test cases to check when our scrapers require updates
- Need sentry logs to detect if we are failing certain entries for certain websites

backend/penndata/management/commands/get_rodin_events.py
judtinzhang: true!

backend/.devcontainer/Dockerfile
```diff
@@ -3,6 +3,7 @@ FROM --platform=linux/amd64 python:3.11-slim-buster
 ARG IMAGE_NAME=pennlabs/mobile-backend-devcontainer
 
 # Install build dependencies
+
```
judtinzhang, line 6: nit: remove extra line

backend/penndata/management/commands/get_wharton_events.py
```diff
@@ -0,0 +1,71 @@
...
+            link = entry.find(class_="entry-title").a["href"]
+
+            info = entry.find(class_="info").span.text.strip()
+            match = re.match(r"(\w+\s+\d+) \| (\d{1,2}:\d{2} [AP]M) - (\d{1,2}:\d{2} [AP]M)", info)
+            if match:
+                _, start_time, end_time = match.groups()
+                start_time_obj = datetime.datetime.strptime(start_time, "%I:%M %p")
+                end_time_obj = datetime.datetime.strptime(end_time, "%I:%M %p")
+            else:
+                match = re.match(
```
judtinzhang, line 40: nit: can we add a comment here to describe what this regex is matching on?

judtinzhang (commented):
LGTM! 2 minor comments; as long as it runs on dev container we should be good.
I'll only this review as comment and @vcai122 can approve it after his review

backend/penndata/management/commands/get_wharton_events.py
```diff
@@ -0,0 +1,73 @@
...
+                    start_time_obj = (
+                        datetime.datetime.strptime(start_time, "%I:%M %p") if start_time else None
+                    )
+                    end_time_obj = (
+                        datetime.datetime.strptime(end_time, "%I:%M %p") if end_time else None
+                    )
+                else:
+                    print("Error: Cannot find date, update scraper.")
+                    return
+            location = ",".join([info.split("•")[-2], info.split("•")[-1]])
```
vcai122, line 58: `location = ",".join(info.split("•")[-2:])`

backend/penndata/management/commands/get_college_house_events.py
```diff
@@ -0,0 +1,131 @@
...
+            if not (url := event_link.get("href")):
+                continue
+            index = calendar_url.find("/", calendar_url.find("://") + 3)
+            base_url = calendar_url[:index]
+            url = f"{base_url}{url}"
+
+            location, start_time, end_time, description, image_url = self.scrape_details(url)
+
+            base_url = "/".join(calendar_url.split("/", 3)[:3])
+            event_type = EVENT_TYPE_MAP.get(base_url, None)
```
vcai122, line 115: why not just pass this in as well?

backend/penndata/management/commands/get_college_house_events.py
```diff
@@ -0,0 +1,131 @@
...
+    "https://lauder.house.upenn.edu": Event.TYPE_LAUDER_COLLEGE_HOUSE,
+    "https://hill.house.upenn.edu": Event.TYPE_HILL_COLLEGE_HOUSE,
+    "https://kcech.house.upenn.edu": Event.TYPE_KCECH_COLLEGE_HOUSE,
+    "https://ware.house.upenn.edu": Event.TYPE_WARE_COLLEGE_HOUSE,
+    "https://fh.house.upenn.edu": Event.TYPE_FH_COLLEGE_HOUSE,
+    "https://riepe.house.upenn.edu": Event.TYPE_RIEPE_COLLEGE_HOUSE,
+    "https://dubois.house.upenn.edu": Event.TYPE_DUBOIS_COLLEGE_HOUSE,
+    "https://gregory.house.upenn.edu": Event.TYPE_GREGORY_COLLEGE_HOUSE,
+    "https://stouffer.house.upenn.edu": Event.TYPE_STOUFFER_COLLEGE_HOUSE,
+}
```
vcai122, line 26: list of pairs makes more sense?

vcai122 (changes requested):
LGTM. requested some small changes but good to go otherwise!

vcai122 (approved):
LGTM

PR #271, opened 2024-03-17
Add Sublet Drafting

dr-Jess:
- Add sublet drafting field
- Implement validation to ensure minimum required fields when publishing


backend/sublet/serializers.py
```diff
@@ -61,6 +61,35 @@ class SubletSerializer(serializers.ModelSerializer):
...
+                or (
+                    field not in validated_data
+                    and (
+                        instance is None
+                        or not hasattr(instance, field)
+                        or not getattr(instance, field)
+                    )
+                )
+            ]:
+                raise serializers.ValidationError(
```
dr-Jess, line 89: Haha this doesn't work I don't know why haha 😭 will fix

judtinzhang: thanks haha. is there any way we can break this down? i get that its the least amount of lines but i feel like its getting a little hard to read

judtinzhang: would really suggest breaking it into first checking if the function came from a PUT/PATCH or a POST

dr-Jess: so it would seem that patching with "" as the address passes the old address into validated_data??? really not sure why but will investigate further tomorrow.

judtinzhang: that is weird ...

backend/sublet/serializers.py
```diff
@@ -61,6 +61,31 @@ class SubletSerializer(serializers.ModelSerializer):
...
+            "price",
+            "negotiable",
+            "start_date",
+            "end_date",
+            "expires_at",
+        ]
+
+        def field_bad(field, validated_data, instance):
+            if field in validated_data:
+                if not validated_data[field]:
```
judtinzhang, line 77: suggest: can compress these two if statements into one

backend/sublet/serializers.py
```diff
@@ -61,6 +61,31 @@ class SubletSerializer(serializers.ModelSerializer):
...
+                    return True
+            else:
```
judtinzhang, line 79: can compress this into `elif`

judtinzhang (commented):
LGTM! Code cleanup comments

backend/sublet/serializers.py
```diff
@@ -61,6 +61,31 @@ class SubletSerializer(serializers.ModelSerializer):
...
+            "negotiable",
+            "start_date",
+            "end_date",
+            "expires_at",
+        ]
+
+        newest_fields = [
+            (field, validated_data[field])
+            if field in validated_data
+            else (field, getattr(instance, field))
```
judtinzhang, line 78: `getattr` has a default parameter, so we can do `getattr(instance, field, None)` to simplify a bit!

backend/sublet/serializers.py
```diff
@@ -61,6 +61,31 @@ class SubletSerializer(serializers.ModelSerializer):
...
+            "title",
+            "address",
+            "price",
+            "negotiable",
+            "start_date",
+            "end_date",
+            "expires_at",
+        ]
+
+        newest_fields = [
```
judtinzhang, line 75: Looks great, a lot simpler and readable! :D

backend/sublet/views.py
```diff
@@ -148,6 +148,7 @@ def list(self, request, *args, **kwargs):
             queryset = queryset.filter(subletter=request.user)
         else:
             queryset = queryset.filter(expires_at__gte=timezone.now())
+            queryset = queryset.filter(is_published=True)
```
judtinzhang, line 151: Nice

judtinzhang (commented):
Almost there! Need one more cycle of reviews because there is currently a test failing. Minor comment

backend/sublet/serializers.py
```diff
@@ -61,6 +61,36 @@ class SubletSerializer(serializers.ModelSerializer):
...
+            (
+                field,
+                (
+                    validated_data[field]
+                    if field in validated_data
+                    else getattr(instance, field, None)
+                    if instance
+                    else None
+                ),
```
vcai122, line 84: technically i think u can also do `validated_data.get(field) or getattr(instance, field, None)`

vcai122 (approved):

backend/sublet/serializers.py
judtinzhang: image


Yep, confirmed! I think I like this better can we perhaps change it to that

PR #274, opened 2024-03-22
WIP: sublet

jonrxu:
for vincent


sublet/components/ui/aspect-ratio.tsx
```diff
@@ -0,0 +1,7 @@
+"use client"
```
vcai122, line 1: is this ui folder completely custom or from somewhere? if its from somewhere else is there anyway we can just yarn install and then import it?

sublet/fonts/satoshi/Satoshi-Black.eot
vcai122, line 1: is there any reason we manually put in the font files? (similar to the ui folder, can we just import the font?)

sublet/public/hamco.jpeg
vcai122, line 1: assuming this is just for testing?

sublet/public/icons/analytics.svg
vcai122, line 1: should we consider using a icon library? (for sake of ease + easier to use in future)

sublet/public/landing-page.jpg
vcai122, line 1: if these landing page images and icons are going to be shared by both sublet and portal we should probably eventually just put them in a shared place so we only have 1 copy

sublet/services/propertyService.ts
```diff
@@ -0,0 +1,154 @@
...
+    for (let i = 0; i < cookies.length; i++) {
+      const cookie = cookies[i].trim();
+      // Does this cookie string begin with the name we want?
+      if (cookie.startsWith(name + '=')) {
+        cookieValue = decodeURIComponent(cookie.substring(name.length + 1));
+        break;
+      }
+    }
+  }
+  return cookieValue;
```
vcai122, line 17: `pref`: avoid for loops and prefer a more functional approach. 
See: https://github.com/pennlabs/penn-mobile/blob/master/frontend/utils/csrf.ts
`pref`: Also prefer f strings over using + for readability

Also this would be one of the things that should be refactored into being shared with portal (prolly also the other auth stuff too)

sublet/services/propertyService.ts
```diff
@@ -0,0 +1,154 @@
...
+  // Make the fetch request with the headers
+  const response = await fetch(url, {
+    method: 'GET',
+    headers: headers
+  });
+
+  // Check if the response was not ok
+  if (!response.ok) {
+    // Attempt to parse the error response body
+    let errorMsg = `Error ${response.status}: ${response.statusText}`;
```
vcai122, line 47: `nit` errorMsg doesn't need to be declared out here. 
Instead just throw the appropriate error inside the try:
`throw new Error('Error ${response.status}: ${response.statusText}: ${errorBody.detail || JSON.stringify(errorBody)}');`

sublet/services/propertyService.ts
```diff
@@ -0,0 +1,154 @@
...
+      // If there's an error parsing the error response, log it
+      console.error('Error parsing the error response:', e);
+    }
+
+    // Throw a new error with the error message
+    throw new Error(errorMsg);
+  }
+
+  const properties: PropertyInterface[] = await response.json();
+  return properties;
```
vcai122, line 61: `nit` no need to set to properties variable if you are going to just return it

sublet/services/propertyService.ts
```diff
@@ -0,0 +1,154 @@
...
+  const response = await fetch(url, {
+    method: 'POST',
+    headers: headers,
+    body: JSON.stringify(property),
+  });
+
+  // Check if the response was not ok
+  if (!response.ok) {
+    // Attempt to parse the error response body
+    let errorMsg = `Error ${response.status}: ${response.statusText}`;
```
vcai122, line 96: same as above

sublet/services/propertyService.ts
```diff
@@ -0,0 +1,154 @@
...
+  // Set up the HTTP headers with Basic Authentication
+  const headers = new Headers({
+    'Authorization': `Basic ${encodedCredentials}`
+  });
+
+  // Make the fetch request with the headers
+  const response = await fetch(url, {
+    method: 'GET',
+    headers: headers
+  });
```
vcai122, line 134: yea you do this alot. Probably best to make a wrapper function that takes in a method and potentially a body. Then it creates the appropriate header based on the method, and attatches the body if needed. 

Because your header and fetch logic is very similar everywhere

sublet/services/propertyService.ts
```diff
@@ -0,0 +1,154 @@
...
+      const errorBody = await response.json();
+      errorMsg += `: ${errorBody.detail || JSON.stringify(errorBody)}`;
+    } catch (e) {
+      // If there's an error parsing the error response, log it
+      console.error('Error parsing the error response:', e);
+    }
+
+    // Throw a new error with the error message
+    throw new Error(errorMsg);
+  }
```
vcai122, line 150: also if you are doing the same error checking each time also best to extract to a function

sublet/utils/csrf.ts
vcai122, line 1: I assume these files are remnants from portal that you copied over?

sublet/app/dashboard/page.tsx
```diff
@@ -0,0 +1,78 @@
...
+import Image from "next/image";
+
+const Dashboard = () => {
+  const [properties, setProperties] = useState<PropertyInterface[]>([]);
+
+  useEffect(() => {
+    fetchProperties()
+      .then((data) => {
+        setProperties(data);
+      })
```
vcai122, line 27: `nit`: `fetchProperties().then(setProperties);` should also work

sublet/components/custom/Property.tsx
```diff
@@ -0,0 +1,169 @@
...
+  const date = new Date(dateString);
+
+  // Format the date to the desired format
+  const formattedDate = date.toLocaleDateString('en-US', {
+    month: 'short', // Abbreviated month name
+    day: 'numeric'  // Numeric day of the month
+  });
+
+  return formattedDate;
+}
```
vcai122, line 65: probably best to have this in a util folder (perhaps globally shared)

sublet/Pipfile
vcai122, line 1: i don't think we need a pipfile here?

vcai122 (commented):
Took a quick glance, overall looks pretty good so far! Make sure to keep in mind what integration/shared code with portal/future projects might look like.

Great work!!

PR #275, opened 2024-03-24
Allow users to see their own expired sublets

dr-Jess:


backend/sublet/views.py
```diff
@@ -151,6 +152,8 @@ def list(self, request, *args, **kwargs):
                 queryset = queryset.filter(amenities__name=amenity)
         if subletter.lower() == "true":
             queryset = queryset.filter(subletter=request.user)
+        else:
+            queryset = queryset.filter(expires_at__gte=timezone.now())
```
vcai122, line 156: nit this logic is slightly different so i would put it on top before all the other ifs.

vcai122 (approved):

PR #279, opened 2024-03-28
Fix Precommit

dr-Jess:
Fixes the pre-commit-config.yaml and adds pre-commit to the pipfile; see [https://pre-commit.com/](url) for usage instructions to NEVER LINT AGAIN :shipit:


dr-Jess (commented):
@judtinzhang can you lock again for me

judtinzhang (approved):

PR #284, opened 2024-04-16
Fix portal homepage logic

vcai122:


frontend/pages/index.tsx
```diff
@@ -23,21 +23,20 @@ export const getServerSideProps = async (
...
-    return {
-      redirect: {
-        destination: `/api/user/clear-cookies/?next=${DASHBOARD_ROUTE}`,
-        permanent: false,
-      },
-    }
+  if (
+    res.status === 403 &&
+    (await res.json()).detail ===
+      'Authentication credentials were not provided.'
```
vcai122, line 29: not too happy about this hard coding but both errors return a 403, need some way to distinguish...

another thought was making a backend route soley to check if the user was logged into our backend, however thought this was overkill...

frontend/pages/index.tsx
```diff
@@ -23,21 +23,20 @@ export const getServerSideProps = async (
...
-    }
-  } else {
-    return {
-      redirect: {
-        destination: `/api/user/clear-cookies/?next=${DASHBOARD_ROUTE}`,
-        permanent: false,
-      },
-    }
+  if (
+    res.status === 403 &&
```
judtinzhang, line 27: idk Javascript but is there a reason you don't need to await for status but do need to for the json?

judtinzhang (approved):
LGTM. My only concern here is now that we are catching the 403 errors on backend, we no longer have the stack trace on sentry. Should be fine though since we already know the root problem + we can test on bastion

frontend/pages/index.tsx
vcai122: no clue tbh, i've just accepted it ig lmao.

vcai122:
> LGTM. My only concern here is now that we are catching the 403 errors on backend, we no longer have the stack trace on sentry. Should be fine though since we already know the root problem + we can test on bastion

we should set up some sort of logging

PR #286, opened 2024-04-19
Upgrade celery, Black

vcai122:
Upgrade black to upgrade celery. Also got really annoyed by the dependency stuff depending on old click figured it was prolly time anyway. 
Minimal changes to actual files (mostly migration file)
We can set up later something to ignore blame on this commit

https://www.stefanjudis.com/today-i-learned/how-to-exclude-commits-from-git-blame/

### DIFF
--- a/backend/Pipfile
+++ b/backend/Pipfile
@@ -4,7 +4,7 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [dev-packages]
-black = "*"
+black = "24.4.0"
 unittest-xml-reporting = "*"
 flake8 = "==6.1.0"
 flake8-isort = "==6.1.0"
--- a/backend/Pipfile
+++ b/backend/Pipfile
@@ -4,7 +4,7 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [dev-packages]
-black = "==19.10b0"
+black = "*"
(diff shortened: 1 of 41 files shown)


backend/dining/migrations/0003_venue_name.py
```diff
@@ -11,6 +11,8 @@ class Migration(migrations.Migration):
 
     operations = [
         migrations.AddField(
-            model_name="venue", name="name", field=models.CharField(max_length=255, null=True),
+            model_name="venue",
+            name="name",
+            field=models.CharField(max_length=255, null=True),
```
vcai122, line 16: new black splits into lines if we have trailing commas, otherwise tries to one line. 
old black tried to one line and added trailing commas. 

To make things look similar for most files where we had one line previously I deleted the trailing comma and reformatted but I ignored for migration files.

backend/Pipfile
```diff
@@ -4,7 +4,7 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [dev-packages]
-black = "==19.10b0"
+black = "*"
```
judtinzhang, line 7: can we fix the black version here?

judtinzhang (approved):
One comment. Also after this is merged in we should remind everyone to update their environment

PR #287, opened 2024-04-21
Add git blame ignore for black upgrade

vcai122:
Everyone needs to run to update their git config file to use the added file:`git config blame.ignoreRevsFile .git-blame-ignore-revs`

Also github should automatically pick up on it


judtinzhang (commented):
yuh

PR #290, opened 2024-04-24
add profanity check for sublets

ashleyzhang01:


backend/sublet/serializers.py
```diff
@@ -91,6 +92,21 @@ class Meta:
...
+        return value
+
+    def validate_description(self, value):
+        # Check for profanity in the description
+        if self.contains_profanity(value):
+            raise serializers.ValidationError("The description contains inappropriate language.")
+        return value
+
+    def contains_profanity(self, text):
+        return predict([text])[0] == 1
```
vcai122, line 108: `nit`: don't need == 1 since 1==True and 0==False

backend/Pipfile
```diff
@@ -46,6 +46,8 @@ python-dateutil = "*"
 selenium = "*"
 webdriver-manager = "*"
 pre-commit = "*"
+profanity-check = "*"
+alt-profanity-check = "*"
```
vcai122, line 50: why do we need both? on alt-profanity-check's website it says: "Alt profanity check is a drop-in replacement of the profanity-check library for the not so well maintained", so it seems like just using alt-profanity-check should be fine.

backend/sublet/serializers.py
```diff
@@ -91,6 +92,21 @@ class Meta:
...
         ]
 
+    def validate_title(self, value):
+        # Check for profanity in the title
+        if self.contains_profanity(value):
+            raise serializers.ValidationError("The title contains inappropriate language.")
+        return value
+
+    def validate_description(self, value):
+        # Check for profanity in the description
```
vcai122, line 102: lmao huge nit but i would argue doesn't need comments since its already readable

vcai122 (changes requested):
mainly just the pipfile comment, but looks pretty good otherwise! 0 shots

backend/Pipfile
ashleyzhang01: i had some weird installation errors without the alt-profanity-check, and it wouldn't run/pipfile wouldn't lock. online fixes said to do this

ashleyzhang01: oh wait jk getting rid of the first ill do that

backend/sublet/serializers.py
```diff
@@ -91,6 +92,19 @@ class Meta:
...
+            raise serializers.ValidationError("The title contains inappropriate language.")
+        return value
+
+    def validate_description(self, value):
+        if self.contains_profanity(value):
+            raise serializers.ValidationError("The description contains inappropriate language.")
+        return value
+
+    def contains_profanity(self, text):
+        return predict([text])[0]
```
judtinzhang, line 106: to confirm, do we need to lowercase the text or anything? or is this package already handle this

judtinzhang (approved):
LGTM! One quick question. Thanks ashley, 0 shots

backend/sublet/serializers.py
ashleyzhang01: it's handled

vcai122 (approved):
LGTM!!!!!!!!!!!!!!!!

PR #291, opened 2024-04-25
Refactor portal serializer + Add email logic for updates on Content

vcai122:
* refactor portal serializer to use abstract class + deal with special case for form data

* set up email configs + email utils (wrapper around django email)

* set up email notifs on content save/update


backend/utils/email.py
```diff
@@ -0,0 +1,33 @@
...
+
+@shared_task(name="utils.send_mail")
+def send_mail(subject, recipient_list, message=None, html_message=None):
+    if recipient_list is None:
+        return False
+    success = django_send_mail(
+        subject=subject,
+        message=message,
+        from_email=None,
+        recipient_list=recipient_list if isinstance(recipient_list, list) else [recipient_list],
```
judtinzhang, line 15: Not a huge fan of this line, especially since the name of the variable has "list" in it. Can we just enforce that `recipient_list` is always of type `list`?

backend/utils/email.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@shared_task(name="utils.send_mail")
+def send_mail(subject, recipient_list, message=None, html_message=None):
+    if recipient_list is None:
+        return False
+    success = django_send_mail(
+        subject=subject,
+        message=message,
+        from_email=None,
```
judtinzhang, line 14: Any reason this is `None`?

backend/utils/email.py
```diff
@@ -0,0 +1,33 @@
+from celery import shared_task
+from django.contrib.auth.models import Group
+from django.core.mail import send_mail as django_send_mail
+from django.template.loader import get_template
+
+
+@shared_task(name="utils.send_mail")
+def send_mail(subject, recipient_list, message=None, html_message=None):
+    if recipient_list is None:
```
judtinzhang, line 9: When will this be the case? I feel like if this is ever `None` we shouldn't fail silently like this (the return value of this function is never checked)

judtinzhang (commented):
Looks good! Some comments and questions

backend/pennmobile/settings/base.py
```diff
@@ -178,3 +178,11 @@
...
 AWS_S3_FILE_OVERWRITE = False
 AWS_DEFAULT_ACL = "public-read"
+
+EMAIL_BACKEND = "django.core.mail.backends.smtp.EmailBackend"
+EMAIL_HOST = os.environ.get("SMTP_HOST", "")
+EMAIL_USE_TLS = True
+EMAIL_PORT = os.environ.get("SMTP_PORT", 587)
+EMAIL_HOST_USER = os.environ.get("SMTP_USERNAME", "")
+EMAIL_HOST_PASSWORD = os.environ.get("SMTP_PASSWORD", "")
+DEFAULT_FROM_EMAIL = os.environ.get("SMTP_FROM_EMAIL", EMAIL_HOST_USER)
```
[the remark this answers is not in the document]
vcai122: host user might be `accounts@pennlabs.org` and default from might be `Penn Mobile <pennmobile@pennlabs.org>`

backend/portal/models.py
```diff
@@ -48,17 +50,52 @@ class Content(models.Model):
...
+                if self.admin_comment and self.status == self.STATUS_REVISION
+                else ""
+            ),
+        )
+
+    def save(self, *args, **kwargs):
+        prev = self.__class__.objects.filter(id=self.id).first()
+        super().save(*args, **kwargs)
+        if prev is None:
+            return self._on_create()
```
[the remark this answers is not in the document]
vcai122: wait yes need for return here because i want need to hard stop since prev is None. 
it was originally 
`self._on_create()
return`

but i changed it to be fancy

backend/portal/serializers.py
```diff
@@ -10,81 +11,69 @@ class Meta:
...
-        # if Poll is updated, then approve should be false
+        # if Content is updated, then approve should be false
+
         if not self.context["request"].user.is_superuser:
-            validated_data["status"] = Poll.STATUS_DRAFT
+            validated_data["status"] = Content.STATUS_DRAFT
         return super().update(instance, validated_data)
 
 
+class PollSerializer(ContentSerializer):
```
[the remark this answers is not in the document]
vcai122: I think its fine, because also this means we don't have to change when we add multiselect. And even if a malicious actor sent us multiselect=True, we need to first approve it and even if we approve it won't do anything cuz the frontend code doesn't use it right now

backend/portal/models.py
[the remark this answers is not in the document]
judtinzhang: Wait then why don't you just have an `elif` for the latter clause?

backend/portal/serializers.py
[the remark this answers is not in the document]
judtinzhang: Sounds good

backend/portal/models.py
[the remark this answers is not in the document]
vcai122: actually yea i could. I did return because i thought i would have more conditions to send an email

backend/portal/serializers.py
```diff
@@ -10,81 +11,69 @@ class Meta:
...
+        ]
+        validated_data["target_populations"] += TargetPopulation.objects.filter(
+            kind__in=auto_add_kind
+        )
+
+        validated_data["creator"] = user
 
         return super().create(validated_data)
 
     def update(self, instance, validated_data):
```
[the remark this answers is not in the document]
vcai122: The serializer will only act on things coming in through the route, so operations directly through the shell or through admin portal does not go through the serializer. So admin comment/priority does not trigger the serializer. 

With regards to club_comment change, I agree, probably don't need to set it to draft but it seems like a bit of an edge case (can't imagine people writing comments after post is approved), and kinda don't want to deal with writing that code + it'll be an extra query to db to get the previous object.

backend/portal/serializers.py
```diff
@@ -223,106 +212,25 @@ def get_image_url(self, obj):
...
             "image_url",
-            "created_date",
-            "start_date",
-            "expire_date",
-            "club_comment",
-            "admin_comment",
-            "status",
-            "target_populations",
         )
-        read_only_fields = ("id", "created_date", "target_populations")
```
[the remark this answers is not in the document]
vcai122: The reason you put target_population as read_only for Posts is because otherwise, the built in serializer tries to validate / set it, but it cannot succesfully validate the data coming from a form causing errors. (hence you marked it as read_only and then manually did it yourself after the fact).

I've tested post with image + non-empty and empty list together... after dealing with the edge case in is_valid in the Post serializer, it should be treating the target_population as any other field now

backend/utils/email.py
vcai122: if we set to None, it uses the default one specified in settings (base.py)

backend/utils/email.py
vcai122: I did this because I thought this could just be "intended bevahior" for this function for sake of ease lol.

A case where we need this right now is if a post without a user attached onto it (so old posts) are updated, the recipient list would be None. 

I guess the real question is whether this should be the intended behavior of this function. I have no preference. We can throw an error instead and do the None checking on the side of the caller of this function instead

backend/portal/models.py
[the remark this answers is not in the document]
judtinzhang: yep up to you

backend/portal/serializers.py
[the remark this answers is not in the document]
judtinzhang: Oh right. Okay yea sounds good for the priority and admin comment. And sure we can keep as is for now

backend/portal/serializers.py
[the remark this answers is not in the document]
judtinzhang: Bet thanks

backend/utils/email.py
judtinzhang: I think I would prefer erroring here. Definitely will save a few minutes if someone needs to debug this code later on

backend/portal/models.py
```diff
@@ -48,17 +50,54 @@ class Content(models.Model):
     admin_comment = models.CharField(max_length=255, null=True, blank=True)
     target_populations = models.ManyToManyField(TargetPopulation, blank=True)
     priority = models.IntegerField(default=0)
+    creator = models.ForeignKey(User, on_delete=models.SET_NULL, null=True, blank=True)
 
     class Meta:
         abstract = True
 
+    def _get_email_subject(self):
```
tuneerroy, line 58: if these are meant to be private methods, they should start with two underscores, not one.

backend/portal/models.py
```diff
@@ -48,17 +50,54 @@ class Content(models.Model):
...
+            self._get_email_subject(),
+            get_backend_manager_emails(),
+            (
+                f"A new {self.__class__._meta.model_name} for {self.club_code}"
+                f"has been created by {self.creator}."
+            ),
+        )
+
+    def _on_status_change(self):
+        if email := getattr(self.creator, "email", None):
```
tuneerroy, line 72: don't need to pass in the last param for `getattr` since the default return value if not found is `None`.

also... should do `if _ is not None` ... more Pythonic :)

backend/portal/serializers.py
```diff
@@ -223,106 +221,25 @@ def get_image_url(self, obj):
...
-            ids = raw_target_populations
-        else:
-            ids = (
-                list()
-                if len(raw_target_populations) == 0
-                else [int(id) for id in raw_target_populations.split(",")]
+    def is_valid(self, *args, **kwargs):
+        if isinstance(self.initial_data, QueryDict):
+            self.initial_data = self.initial_data.dict()
+            self.initial_data["target_populations"] = list(
```
tuneerroy, line 238: hot damn, a `list` and then a `map`...

backend/portal/serializers.py
```diff
@@ -223,106 +221,25 @@ def get_image_url(self, obj):
...
+                (
+                    map(int, self.initial_data["target_populations"].split(","))
```
tuneerroy, line 240: could also probably just do 
```
self.initial_data["target_population"] = [int(v) for v in self.initial_data.get("target_populations", "").split(",")]
```

tuneerroy (approved):
didn't test it but your test cases look comprehensive so i trust it! ow lgtm :<

backend/portal/models.py
vcai122: no u need the default else it errors. 

but we have this pattern everywhere in our code...

backend/portal/models.py
vcai122: hm
https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name

backend/portal/models.py
tuneerroy: o u right

argh fine keep it

backend/portal/models.py
tuneerroy: damn u right, that's so weird, I've always assumed the name mangling is a good way to enforce private

backend/portal/models.py
```diff
@@ -48,17 +50,54 @@ class Content(models.Model):
     admin_comment = models.CharField(max_length=255, null=True, blank=True)
```
minghansun1, line 50: In line 50, created_date should probably be immutable. 
created_date = models.DateTimeField(default=timezone.now, editable=False)

Update save to
def save(self, *args, **kwargs):
    if not self.pk:
        self.created_date = timezone.now()  # Set created_date to the current time
    else:
        if 'update_fields' in kwargs and 'created_date' in kwargs['update_fields']:
            kwargs['update_fields'].remove('created_date')

    prev = self.__class__.objects.filter(id=self.id).first()
    super().save(*args, **kwargs)

    if prev is None:
        self._on_create()
        return

    if self.status != prev.status:
        self._on_status_change()

backend/portal/serializers.py
```diff
@@ -10,81 +11,78 @@ class Meta:
...
             "admin_comment",
             "status",
             "target_populations",
         )
         read_only_fields = ("id", "created_date")
+        abstract = True
+
+    def _auto_add_target_population(self, validated_data):
+        # auto add all target populations of a kind if not specified
+        if target_populations := validated_data.get("target_populations"):
```
minghansun1, line 32: May be more efficient to do this using set difference instead of nested loops

minghansun1 (commented):
General feedback: Everything looks logical to me, but I'm not familiar with Django best practices so it's hard for me to find specific issues.

Design Consideration:
For PollVotes, we could give users the option to be anonymous or named, and let us filter results based anonymous votes, named votes, and overall votes.

Specific Issues: Make created_date immutable, do _auto_add_target_population using set difference instead of nested loops

backend/portal/models.py
vcai122: If you look in the serializer for content, you'll see that created_date is read only so it is enforced there. 
Technically an admin could change the created date, but there is no way to prevent this since you can always change a row in a db. In other words you can't enforce atomicity at the db level

backend/portal/models.py
vcai122: wait i didn't say i preferred single underscore. tbh i don't know which i prefer was hoping for u to choose one lol.

backend/portal/serializers.py
vcai122: Good thought, although it doesn't matter here since if you look Kind has like 4 values. Probably the overhead from sets is more costly anyway

backend/portal/models.py
tuneerroy: lol i'd just stick with SO's opinion, single underscore for private methods, double underscore if you're concerned about inheritance overriding

PR #292, opened 2024-04-26
Introducing Analytics!

judtinzhang:
- NOTE: to run locally, must add `CLIENT_ID`, `CLIENT_SECRET`, and `PLATFORM_URL` to environment


backend/pennmobile/analytics.py
```diff
@@ -0,0 +1,15 @@
...
+
+AnalyticsEngine = LabsAnalytics()
+
+
+class Metric(str, Enum):
+    GSR_BOOK = "gsr.book"
+
+
+def record_analytics(metric: Metric, username=None):
+    txn = AnalyticsTxn(Product.MOBILE_BACKEND, username, data=[{"key": metric, "value": "1"}])
```
judtinzhang, line 14: I'll fix this later. Theoretically the `value` shouldn't matter for the `GSR_BOOK` metric

PR #302, opened 2024-08-04
Fix news fetch

vcai122:


backend/penndata/views.py
```diff
@@ -36,22 +36,30 @@ class News(APIView):
...
         except ConnectionError:
             return None
 
         html = resp.content.decode("utf8")
 
         soup = BeautifulSoup(html, "html5lib")
 
-        frontpage = soup.find("div", {"class": "col-lg-6 col-md-5 col-sm-12 frontpage-carousel"})
+        if not (
+            frontpage := soup.find(
```
ashleyzhang01, line 52: 🦭

backend/portal/management/commands/polls_populate.py
```diff
@@ -73,7 +73,7 @@ def handle(self, *args, **kwargs):
             user_nursing = User.objects.get(username="user_nursing")
```
vcai122, line 74: @tuneerroy Do we know what this command is for

PR #303, opened 2024-08-30
Fix college_house_events request

dr-Jess:


ashleyzhang01 (approved):
what's the reason we need to disable SSL certificate verification? putting a little description in your PRs would be helpful in the future. otherwise looks good 👨‍💻

tuneerroy (changes requested):
Add a comment to the code that explains why we're disabling the extra protection. +1 what Ashley said on the PR description (maybe add an example of a situation that we are now exposed to?)

PR #304, opened 2024-09-01
Jsonify nutrition info

vcai122:


backend/dining/serializers.py
```diff
@@ -10,10 +12,18 @@ class Meta:
 
 
 class DiningItemSerializer(serializers.ModelSerializer):
+    nutrition_info = serializers.SerializerMethodField()
+
     class Meta:
         model = DiningItem
         fields = "__all__"
 
+    def get_nutrition_info(self, obj):
```
tuneerroy, line 21: is this function supposed to be used anywhere? or does it autooverload getting the nutrition_info in the serializer?

tuneerroy (approved):

PR #307, opened 2024-09-27
Make GSR Group Admin easy to add/remove users

vcai122:
For easy adding/removing users to groups

### DIFF
--- a/backend/gsr_booking/admin.py
+++ b/backend/gsr_booking/admin.py
@@ -3,6 +3,27 @@ from django.contrib import admin
 from gsr_booking.models import GSR, Group, GroupMembership, GSRBooking, Reservation
 
 
+class GroupMembershipInline(admin.TabularInline):
+    model = GroupMembership
+    extra = 0
+
+    readonly_fields = ["name"]
+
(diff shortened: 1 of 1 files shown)


backend/gsr_booking/admin.py
```diff
@@ -3,6 +3,29 @@
...
 
 
+class GroupMembershipInline(admin.TabularInline):
+    model = GroupMembership
+    extra = 0
+
+    readonly_fields = ["name"]
+
+    def name(self, obj):
+        return obj.user.get_full_name()
```
dr-Jess, line 13: Probably fine but what if we made this more verbose for better searching/viewing? We only use Labs group at the moment but what if we displayed both username and group name for viewing and searching.

dr-Jess (changes requested):

backend/gsr_booking/admin.py
vcai122: This is inline for a group. AKA the group info will be at the top of the page when anyone is looking here

dr-Jess (approved):
lgtm

PR #308, opened 2024-09-29
[Fix] Laundry Rewrite to Account For New SpeedQueen API

dr-Jess:
Rewrite laundry to handle Penn's new SpeedQueen API, without changing anything that would affect the frontend. Future plans are to hard rewrite (new API has more data, such as whether a machine door is open/could be used to tell when laundry hasn't been taken out!!!), but this is a quick fix.

Some notes on the new API:
Statuses are entirely reworked, but I've left the possible options as the old ones and lined them up best I can as to not break the frontend.
Additionally, the API is now actually relational!! with locations and rooms instead of a page to be scraped. This means we'll have to properly track both locations and the rooms in them now, modifying our requests to match this pattern.


dr-Jess:
Currently fails all tests because tests are written on a mock of the old API, will have to rewrite (or maybe we just test it live?) >:)

backend/laundry/api_wrapper.py
```diff
@@ -1,46 +1,58 @@
...
 from laundry.models import LaundryRoom, LaundrySnapshot
 
 
-HALL_URL = f"{settings.LAUNDRY_URL}/?location="
+def get_room_url(room_id: int):
+    return f"{settings.LAUNDRY_URL}/rooms/{room_id}/machines?raw=true"
 
 
-def update_machine_object(cols, machine_object):
+headers = {
```
tuneerroy, line 13: make this a constant `LAUNDRY_HEADERS` or something, you can then import this in your command as well

backend/laundry/api_wrapper.py
```diff
@@ -1,46 +1,58 @@
...
+            machine_type_data["time_remaining"].append(int(time_remaining))
         except ValueError:
             pass
-    elif cols[2].getText() == "Out of order":
-        machine_object["out_of_order"] += 1
-    elif cols[2].getText() == "Not online":
-        machine_object["offline"] += 1
+    elif status in ["AVAILABLE", "COMPLETE"]:
+        machine_type_data["open"] += 1
+    # elif status == "Out of order":
```
tuneerroy, line 36: let's remove this dead code

backend/laundry/management/commands/load_laundry_rooms.py
```diff
@@ -8,19 +8,20 @@
 class Command(BaseCommand):
     def handle(self, *args, **kwargs):
 
-        with open("laundry/data/laundry_data.csv") as data:
+        with open("laundry/data/laundry_data_new.csv") as data:
```
tuneerroy, line 11: let's rename the file to `laundry_data.csv`?

backend/laundry/management/commands/update_laundry_rooms.py
```diff
@@ -0,0 +1,66 @@
...
+                )
+                room_request.raise_for_status()
+            except HTTPError as e:
+                self.stdout.write(f"Error: {e}")
+                return
+
+            room_request_json = room_request.json()
+            # count washers and dryers
+            count = (0, 0)
+            f = lambda acc, x: (
```
tuneerroy, line 55: this is unreadable dude, just make it a for loop

backend/laundry/management/commands/update_laundry_rooms.py
```diff
@@ -0,0 +1,66 @@
...
+                (acc[0] + 1, acc[1])
+                if x["isWasher"]
+                else (acc[0], acc[1] + 1) if x["isDryer"] else acc
+            )
+            count = reduce(f, room_request_json, count)
+            room.extend(count)
+
+        # write to csv
+        with open("laundry/data/laundry_data_new.csv", "w") as data:
```
tuneerroy, line 64: rename `data` to `f`. it's a file ptr not data content. usually `data = f.read()`.

backend/laundry/management/commands/update_laundry_rooms.py
```diff
@@ -0,0 +1,66 @@
...
+
+        except HTTPError as e:
+            self.stdout.write(f"Error: {e}")
+            return
+
+        all_rooms_request_json = all_rooms_request.json()
+        locations = all_rooms_request_json["geoBoundaries"][0]["geoBoundaries"]
+
+        laundry_rooms = [
+            [room["id"], room["roomName"], location["description"], location["id"]]
```
tuneerroy, line 33: use a dictionary, not a list of lists, like
```
{
  "room_id": ...
   ...
}
```

tuneerroy (changes requested):
just did a brief glance, dk about logic

tuneerroy:
> Currently fails all tests because tests are written on a mock of the old API, will have to rewrite (or maybe we just test it live?) >:)

sorry kiddo, gotta write the tests before we deploy :)

backend/laundry/models.py
```diff
@@ -5,18 +5,18 @@
 
 
 class LaundryRoom(models.Model):
```
vcai122, line 7: Can you explain why previously we had a hall_id and why we have a room_id now? 

What is the difference between them? 

Depending on this answer, I may be hesitant to delete UUID.

vcai122:
> Currently fails all tests because tests are written on a mock of the old API, will have to rewrite (or maybe we just test it live?) >:)

Have you tested this manually? 

I think it wouldn't be too much of a lift to mock the new one? It should be just in a couple places right?

backend/laundry/management/commands/update_laundry_rooms.py
vcai122: ifl u could also just put the room object itself

backend/laundry/management/commands/update_laundry_rooms.py
```diff
@@ -0,0 +1,66 @@
+import csv
```
vcai122, line 1: what is this new file for? Can you describe the changes between the new API and old API in the PR description for more context about what you are doing?

backend/laundry/management/commands/update_laundry_rooms.py
vcai122: but yea lets not have indexes be symbolic lets be explicit

backend/laundry/management/commands/update_laundry_rooms.py
vcai122: no just do it seperately with two sums...

backend/laundry/api_wrapper.py
```diff
@@ -49,41 +61,28 @@ def parse_a_hall(hall_link):
...
+            # TODO: Unsure if remaining seconds and the new status ids work on frontend
+            detailed.append(
+                {
+                    "id": machine["id"],
+                    "type": "washer" if machine["isWasher"] else "dryer",
+                    "status": machine["currentStatus"]["statusId"],
+                    "time_remaining": machine["currentStatus"]["remainingSeconds"],
+                }
+            )
     return {"washers": washers, "dryers": dryers, "details": detailed}
```
vcai122, line 86: We should make this more readable. Compute washers, dryers, and detailed seperately. Not at the same time.

vcai122 (changes requested):
HE DID IT!!!!!!! 6 shots

left most of my questions throughout

couple notes for READABILITY
- indexes are EVIL
- compute stuff seperately if it makes it more clear (don't jam everything in the same for loop this is not a leetcode problem, but also you shouldn't be solving ur leetcode problems that way anyway)
- similar to above don't feel the need to be ultra efficient by only doing 1 for loop. Readability is number 1 concern

backend/laundry/api_wrapper.py
dr-Jess: I left this here so we know what some of the legacy code does. Want to full rewrite at some point, but this is in case we find out these statuses work with the new machines too.

backend/laundry/models.py
dr-Jess: hall_id was just a generic id/pkey (0, 1, 2...) on our end. UUID was a specific id that was used to access the room from an end point.

The new API has a more relational structure, where there's a one to many between locations and rooms. To keep track of this, we need to store both.

dr-Jess:
Made significant changes + testing; before merging please don't allow me to forget to squash migrations. How many shots?

backend/laundry/api_wrapper.py
```diff
@@ -92,26 +101,10 @@ def check_is_working():
...
-            },
-        )
-        r.raise_for_status()
-        return (
-            "The transaction log for database 'QuantumCoin' is full due to 'LOG_BACKUP'."
-            not in r.text
-        )
-    except HTTPError:
+    all_rooms_request = get_validated(f"{settings.LAUNDRY_URL}/geoBoundaries/5610?raw=true")
+    if all_rooms_request is None:
```
tuneerroy, line 105: lets just do `return all_rooms_request is not None`

backend/laundry/serializers.py
```diff
@@ -4,6 +4,9 @@
 
 
 class LaundryRoomSerializer(serializers.ModelSerializer):
+    # tech debt but we're rewriting at some point and this maintains frontend-compatibility
+    hall_id = serializers.IntegerField(source="room_id")
```
tuneerroy, line 8: should we create an issue for this?

backend/penndata/views.py
```diff
@@ -174,10 +174,11 @@ def get(self, request):
 
         cells = []
 
-        # adds laundry preference to home, defaults to Bishop if no preference
+        # adds laundry preference to home, defaults to 0 if no preference
+        # TODO: This defaults to a nonexistent room, not sure if that's problematic
```
tuneerroy, line 178: is there a way we can log this so if this does happen, we can see it?

backend/laundry/serializers.py
dr-Jess: Yeah made

vcai122 (approved):
LGTM
i think?

tuneerroy (approved):
LGTM!

PR #309, opened 2024-09-29
Created models for penn wrapped -- issue with pip files though

zachHarpaz:


backend/wrapped/models.py
```diff
@@ -0,0 +1,76 @@
...
+    class Meta:
+        unique_together = ("stat_key", "semester")
+
+    def __str__(self):
+        return f"Global -- {self.stat_key}-{str(self.semester)} : {self.stat_value}"
+
+
+# Add a new model for keys 
+class UserStatKey(models.Model):
+    stat_key = models.CharField(max_length=50, null=False, blank=False)
```
vcai122, line 28: same here about `stat_` prefix. Also make this the primary key

backend/wrapped/models.py
```diff
@@ -0,0 +1,76 @@
...
+    def __str__(self):
+        return f"{self.text_field_name} -> {self.IndividualStatKey} --> {self.GlobalStatKey}"
+
+# Why do we want time? Ask Vincent 
+class Page(models.Model):
+
+    # How to do this, using individual vs stat_key ?
+    name = models.CharField(max_length=50, null=False, blank=False)
+    template_path = models.CharField(max_length=50, null=False, blank=False)
+    stat_locations = models.ManyToManyField(StatLocation, blank=True)
```
vcai122, line 73: let's make this a through model see Gsr Group to User through GroupMembership as an example.

Also let's not make this "one or the other" model. Let's just make the administrator specify whether it would be individual or global through 2 seperate many to manys. It seems reasonable for the administrator to think about this anyway.

backend/wrapped/models.py
```diff
@@ -0,0 +1,76 @@
...
+
+        super(StatLocation, self).save(*args, **kwargs)
+    def __str__(self):
+        return f"{self.text_field_name} -> {self.IndividualStatKey} --> {self.GlobalStatKey}"
+
+# Why do we want time? Ask Vincent 
+class Page(models.Model):
+
+    # How to do this, using individual vs stat_key ?
+    name = models.CharField(max_length=50, null=False, blank=False)
```
vcai122, line 71: This name should be for internal use only. Also make it a primary key to ensure uniqueness.

backend/wrapped/views.py
```diff
@@ -0,0 +1,36 @@
...
+from django.contrib.auth import get_user_model
+from rest_framework import generics, viewsets
+from rest_framework.decorators import action
+from django.shortcuts import get_object_or_404
+from .serializers import PageSerializer
+
+User = get_user_model()
+
+
+class GeneratePage(viewsets.ModelViewSet):
```
vcai122, line 17: get the user through `request.user`, and make sure you have to be authenticated to use this route. see our other views.

Also I don't think ModelViewSet is the right viewset to use here.

Also call this and associated urls `Pages`. We're not really generating anything (generating implies side effects). We're just serializing data already in the db: we are returning `Page`s filtered for the given semester, serialized in a particular way.

backend/wrapped/views.py
```diff
@@ -0,0 +1,36 @@
...
+
+        # Find the user
+        user = get_object_or_404(User, username=username)
+
+        # Filter pages by this user
+        # pages = Page.objects.filter(user=user)
+        pages = Page.objects.filter(semester=semester)
+
+        # Serialize the data
+        serializer = PageSerializer(pages, many=True)
```
vcai122, line 35: what does the many=True here do

backend/wrapped/models.py
```diff
@@ -0,0 +1,76 @@
+from django.db import models
```
vcai122, line 1: I know we talked about this before but I think we should actually make Semester a model again.
Reason I'm saying this is for admin page purposes, we can use inline models to manage all the pages for a particular semester at a time which seems like a clean way to do it.

We would make a semester model with 1 field, and make it many to many on pages.

We should think of a Page from now on as a page reusable across semesters

vcai122: Leaving this here because we should probably use this later: https://django-admin-sortable2.readthedocs.io/en/latest/

vcai122: So you have to also make GlobalStatKey model

backend/wrapped/serializers.py
```diff
@@ -0,0 +1,65 @@
...
+    class Meta: 
+        model = GlobalStat
+        fields = ["stat_key", "stat_value", "semester"]
+
+class UserStatKeySerializer(serializers.ModelSerializer):
+    class Meta:
+        model = UserStatKey
+        fields = ["stat_key"]
+
+class PageSerializer(serializers.ModelSerializer):
```
vcai122, line 34: You're not really using the other serializers since you're just serializing it and immediately reformatting it, so might as well not use it. 

The way you should be thinking about serializers is that everytime you have a queryset and then you say to yourself "I wish I could have this queryset in this particular format" you should make a serializer.

backend/wrapped/views.py
vcai122: Actually now that we are adding a Semester model, this should be thought of as returning data associated with 1 particular semester. 

(And in the serializer for semester it turns it into a list of pages each of which can be serialized with the page serializer and so on...)

vcai122 (changes requested):
Good work so far! Django has a lot of intricacies but you're doing a good job navigating them so keep up the good work. Let me know if you have any questions about comments

backend/wrapped/views.py
```diff
@@ -0,0 +1,16 @@
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.views import APIView
+from rest_framework.response import Response
+from rest_framework import status
+from .models import Page, Semester
+from .serializers import PageSerializer
+from .serializers import SemesterSerializer
+
```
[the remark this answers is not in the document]
vcai122: Note you don't need to pass in user in context. See how we access the user in serializers in our other code (e.g. Portal serializer)

backend/wrapped/serializers.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class SemesterSerializer(serializers.ModelSerializer):
+    pages = serializers.SerializerMethodField()
+
+    class Meta:
+        model = Semester
+        fields = ['semester', 'pages']
+
+    def get_pages(self, obj):
```
vcai122, line 100: You are manually writing what django can already do if you just hook up pages directly with the serializer. (See sublet serializers for example). 
e.g.
```class SubletSerializerRead(serializers.ModelSerializer):
    amenities = serializers.PrimaryKeyRelatedField(
        many=True, queryset=Amenity.objects.all(), required=False
    )
    images = SubletImageURLSerializer(many=True, required=False)
```

backend/wrapped/serializers.py
```diff
@@ -0,0 +1,105 @@
+from rest_framework import serializers
+from .models import Page, IndividualStat, GlobalStat, IndividualStatThrough, GlobalStatThrough, Semester, User
+
+class IndividualStatSerializer(serializers.ModelSerializer):
+    key = serializers.StringRelatedField()
```
vcai122, line 5: Let's not use StringRelatedField because it relies on the `__str__` which is ususally used for testing. Can you find a better one to use?

backend/wrapped/serializers.py
vcai122: wait sorry i think for this you need to manually pass the semester obj so yours is correct. But its kinda verbose why not just: 

`return PageSerializer(obj.pages.all(), many=True, context={'semester': obj}).data`

backend/wrapped/serializers.py
```diff
@@ -0,0 +1,105 @@
...
+
+    class Meta:
+        model = Page
+        fields = ['name', 'template_path', 'combined_stats']
+
+    def get_combined_stats(self, obj):
+        user = self.context.get('user')
+        semester = self.context.get('semester')
+
+        individual_throughs = IndividualStatThrough.objects.filter(Page=obj)
```
vcai122, line 73: You a bit too much manual serialization here. As an outline the following is a bit more reasonable:
1. Serialize IndividualEntries to become a list of (field_name, value) pairs (using a serializer)
2.Serialize IndividualEntries to become a list of (field_name, value) pairs (using a serializer)
3. Combine into a dict and return. 

Although this seems hard to describe over a code review. Maybe we can sit down and peer program for this part

backend/wrapped/serializers.py
```diff
@@ -0,0 +1,105 @@
...
+
+    class Meta:
+        model = IndividualStatThrough
+        fields = ['text_field_name', 'individual_stat_value']
+
+    def get_individual_stat_value(self, obj):
+        user = self.context.get('user')
+        semester = self.context.get('semester')
+
+        try:
```
vcai122, line 31: prefer filter(...).first() over get()

vcai122 (changes requested):
Looking very promising! Next steps would be to clean up a lot of the dead code and get it ready to merge. 

Most comments I left should be pretty quick and so let's just get this into a ready to merge state soon.

vcai122 (approved):
LGTM! 
Notes before merging:
- Merge in master to get rid of build errors
- Make sure everything is passing (except for test coverage)
- Note down the decrease in code coverage, you'll be responsible for bringing it up in ur next PR

PR #311, opened 2024-10-13
fix(styling): navigation + buttons

jonrxu:
- [x] Some styling changes for navigation and buttons, and other small fixes.
- [x] Replacing outdated CSS with new Tailwind styling.

Screenshot 2024-10-13 at 2 34 20 PM


frontend/components/dashboard/DashboardHeader.tsx
```diff
@@ -42,11 +42,16 @@ export const DashboardHeader = ({
               : CREATE_POLL_ROUTE
           }
         >
-          <a>
-            <Button color={colors.GREEN} round style={{ margin: '0 0 0 1rem' }}>
+          <Button
+            className="flex items-center justify-center"
+            color={colors.GREEN}
```
vcai122, line 47: where is colors coming from? Should add the colors we are using to the tailwind config so we can just use them directly in the classname?

frontend/components/form/StatusBar.tsx
```diff
@@ -43,7 +43,11 @@ const StatusBar = ({ status }: iStatusBarProps) => {
 
   return (
     <>
-      <Group horizontal style={{ margin: '0 0.5rem' }}>
+      <Group
+        horizontal
+        className="flex items-center"
+        style={{ margin: '0 0.5rem' }}
+      >
```
vcai122, line 50: why not use tailwind to do the margin?

frontend/components/styles/Buttons.tsx
```diff
@@ -44,20 +45,29 @@ export const Toggle = s.div`
...
+export const ToggleOption = React.forwardRef<
+  HTMLButtonElement,
+  { active: boolean } & React.ButtonHTMLAttributes<HTMLButtonElement>
+>(({ active, className, ...props }, ref) => (
+  <button
+    ref={ref}
+    type="button"
+    className={`
+      border-0
+      ${active ? 'bg-blue-500 text-white' : 'bg-neutral-100 text-gray-700'}
```
vcai122, line 57: again lets be consistent with our colors. I think we should just take the colors. Either we use colors.whatever or only tailwind. Probably prefer adding the colors we need to the config?

frontend/components/styles/Buttons.tsx
```diff
@@ -66,7 +76,7 @@ export const PostPollToggle = ({
   activeOption: PageType
   setActiveOption: React.Dispatch<React.SetStateAction<PageType>>
 }) => (
-  <Toggle style={{ verticalAlign: 'middle' }}>
+  <div className="flex items-center rounded-full bg-neutral-100 p-1.5 h-10 align-middle border shadow-sm shadow-neutral-50">
```
vcai122, line 79: why replace with div?

frontend/components/styles/Nav.tsx
```diff
@@ -1,68 +1,37 @@
...
-            {user.first_name[0].toUpperCase()}
-            {user.last_name[0].toUpperCase()}
-          </InlineText>
-        </ProfilePicWrapper>
-        <Text bold heading>
-          {user.first_name} {user.last_name}
-        </Text>
-      </Group>
-    </ProfileWrapper>
+    <div className="flex items-center gap-2 font-work-sans">
```
vcai122, line 23: was there a reason you didn't fully translate everything like height adn background color for example?

frontend/components/styles/Nav.tsx
```diff
@@ -72,37 +41,62 @@ const NavItem = ({
...
-  position: fixed;
-`
-
 export const Nav = () => {
   const { user } = useContext(AuthUserContext)
 
   return (
-    <NavWrapper>
+    <div
+      className={`w-[12rem] bg-${colors.NAV_BACKGROUND} flex flex-col justify-between items-center min-h-screen fixed py-4`}
```
vcai122, line 80: same thing about color here? 

also why not move `fixed` over and `text-align`
also why was before min-height 99vh? 
answer could just be you have no clue but the new thing works

frontend/components/styles/sizes.ts
```diff
@@ -1,5 +1,5 @@
 export const NAV_HEIGHT = '4.25rem'
-export const NAV_WIDTH = '14%'
+export const NAV_WIDTH = '12rem'
 export const MAX_BODY_HEIGHT = `calc(100vh - ${NAV_HEIGHT})`
```
vcai122, line 3: think NAV_WIDTH is unused now. But also should we even have this `sizes.ts`? Seems wrong to me given we r using tailwind now?

vcai122 (changes requested):
Also a more general question. If we are revamping? Should we still keep the styled components such as `Button`, `Link`, etc.?

frontend/components/dashboard/DashboardHeader.tsx
jonrxu: Yeah I am planning on this eventually, right now it's more of a gradual adoption of Tailwind, so I want to get everything in place first.

frontend/components/form/StatusBar.tsx
jonrxu: Sounds good. A lot of stuff is just temporary changes while I move everything to Tailwind.

frontend/components/styles/Buttons.tsx
jonrxu: Sounds good.

frontend/components/styles/Buttons.tsx
jonrxu: Because Toggle is nothing more than a div wrapper itself. The whole component should be called Toggle tbh

jonrxu: The CSS for this frontend repo is very very granular, so every div kind of has its own CSS + class

frontend/components/styles/Nav.tsx
jonrxu: A lot of these things are purely aesthetic changes - mostly for styling consistency. I want to make visual changes just to test things first.

frontend/components/styles/Nav.tsx
jonrxu: I think the 99vh was supposed to give a little bit of wiggle room for spacing, and it didn't matter because everything was bg-white.

Eventually, I should add in a plugin that will sort Tailwind classes automatically upon save.

frontend/components/styles/sizes.ts
jonrxu: Yeah, we shouldn't have this anymore, this is more of a gradual change. I might still keep a constants file for various purposes.

vcai122 (approved):

PR #312, opened 2024-10-13
feat(sublet): add FE apis and interfaces

jonrxu:
Add frontend API methods and interfaces for subletting.


frontend/api/sublet/offersApi.ts
```diff
@@ -0,0 +1,39 @@
...
+    throw new Error('Failed to fetch offers')
+  }
+  return response.json()
+}
+
+// Create a new offer for a sublet
+export const createOffer = async (
+  subletId: string,
+  data: CreateOfferRequest
+): Promise<void> => {
```
vcai122, line 20: Probably the response shouldn't be void? We should return the response so we can check it and maybe display an error or sucess message based off of it. 

In general is it nescecary to throw after each call (i have no preference here but just wondering). Why not just return the response and have the caller inspect it? It seems natural that for most responses we might want to handle the happy case and the bad case

PR #316, opened 2024-10-23
Updated models.py

minghansun1:


backend/market/models.py
```diff
@@ -8,51 +8,74 @@
...
+    start_date = models.DateTimeField()
+    end_date = models.DateTimeField()
 
 
-class SubletImage(models.Model):
-    sublet = models.ForeignKey(Sublet, on_delete=models.CASCADE, related_name="images")
-    image = models.ImageField(upload_to="sublet/images")
+class ItemImage(models.Model):
+    item = models.ForeignKey(Item, on_delete=models.CASCADE, related_name="images")
+    image = models.ImageField(upload_to="marketplace/images")
```
vcai122, line 81: add a TODO here so we are reminded to verify this in production

backend/market/models.py
```diff
@@ -8,51 +8,74 @@
...
     message = models.CharField(max_length=255, blank=True)
     created_date = models.DateTimeField(auto_now_add=True)
 
     def __str__(self):
-        return f"Offer for {self.sublet} made by {self.user}"
+        return f"Offer for {self.item} made by {self.user}"
 
 
-class Amenity(models.Model):
+class Category(models.TextChoices):
```
vcai122, line 24: Let's make this an actual model. That way I can add more categories on admin site.

vcai122 (changes requested):
Great start. Just be careful to make sure you aren't missing the small things from before when you rewrite the code! Looking forward to the rest of the changes!

backend/market/models.py
```diff
@@ -8,51 +8,74 @@
...
     def __str__(self):
         return self.name
 
 
-class Sublet(models.Model):
-    subletter = models.ForeignKey(User, on_delete=models.CASCADE)
-    sublettees = models.ManyToManyField(
-        User, through=Offer, related_name="sublets_offered", blank=True
+class Item(models.Model):
+    seller = models.ForeignKey(User, on_delete=models.CASCADE)
```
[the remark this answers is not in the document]
minghansun1: I originally removed it since I thought that it was unnecessary since Offer is an explicitly defined intermediate table for that relation. However, I think I'll add it back since it makes it easier to directly make queries to the buyers of an item.

backend/market/models.py
minghansun1: What exactly do you want us to verify? Is it the image routes?
So i can leave a more specific TODO

backend/market/models.py
```diff
@@ -8,51 +8,74 @@
...
+    tags = models.ManyToManyField(Tag, related_name="items", blank=True)
```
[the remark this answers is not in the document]
minghansun1: If we don't specify related_name, django automatically generates a related_name right?
Also, do you mean we don't need to specify related_name for most things unless there are multiple relationships with the same model in a model?

minghansun1:
Thanks for the feedback!

backend/market/models.py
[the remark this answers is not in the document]
vcai122: Sounds good. Yea, using "through" models is always technically redundant so the reason its used is for sake of ease for syntax

backend/market/models.py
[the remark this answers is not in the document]
vcai122: Yes, django automatically generates it. For the most part we only specify it if there is a strong reason to, one example is

if model A has
- many to many to B
- many to many to B

Then we have to name the relationship specifically, so if we are on B looking at A we know which one we are talking about

backend/market/models.py
vcai122: make sure that this s3 bucket exists/if we need to make it manually or django will make it for us

backend/market/models.py
```diff
@@ -8,51 +8,81 @@
 
 class Offer(models.Model):
     class Meta:
-        constraints = [models.UniqueConstraint(fields=["user", "sublet"], name="unique_offer")]
+        constraints = [models.UniqueConstraint(fields=["user", "item"], name="unique_offer_market")]
 
-    user = models.ForeignKey(User, on_delete=models.CASCADE, related_name="offers_made")
-    sublet = models.ForeignKey("Sublet", on_delete=models.CASCADE, related_name="offers")
+    user = models.ForeignKey(User, on_delete=models.CASCADE, related_name="offers_made_market")
```
vcai122, line 13: are these `_market` because sublet exists? If so let's make a note to fix this after we delete sublet stuff

backend/market/models.py
```diff
@@ -8,51 +8,81 @@
...
     created_date = models.DateTimeField(auto_now_add=True)
 
     def __str__(self):
-        return f"Offer for {self.sublet} made by {self.user}"
+        return f"Offer for {self.item} made by {self.user}"
 
 
-class Amenity(models.Model):
+class Category(models.Model):
+    '''
```
vcai122, line 25: nit: remove this comment. feel free to post it in our slack chat or smth if u want to keep it around somewhere but lets not have dead code

vcai122 (changes requested):
I looked at the models again but just skimmed the serializers. 
I noticed some comments were unaddressed, I think a good system would be if you thumbs up the ones that are addressed (and I can verify and close the conversation). That way we know whats left to address.

backend/market/models.py
```diff
@@ -8,51 +8,74 @@
...
-    end_date = models.DateField()
 
     def __str__(self):
-        return f"{self.title} by {self.subletter}"
+        return f"{self.title} by {self.seller}"
+
+
+class Sublet(models.Model):
+    item = models.OneToOneField(Item, on_delete=models.CASCADE, related_name="sublet")
+    address = models.CharField(max_length=255)
```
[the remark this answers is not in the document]
minghansun1: What do you mean by this? Are you asking me to remove related_name again?

backend/market/models.py
minghansun1: I'll just remove related_name since there's only one ForeignKey to User in Offer.

backend/market/models.py
minghansun1: I kept it in so we remember which Categories to create when deploying the app. But yeah keeping it in Slack will be cleaner.

PR #317, opened 2024-11-05
Add type safety and type hinting

ashleyzhang01:
Type safety is important for catching bugs early, facilitating refactoring, and improving overall maintainability. 
We should use type hints makes the code more understandable to both developers and static type checkers, so our code can be more reliable and predictable

We should:
- Use type hints
- Write code in a more functional style
- Add comments for things like magic values, unintuitive behaviors, gotchas, external context, etc.
- Types should have capitalized names
- Keep things pythonic (e.g. use list comprehensions, avoid explicit for-loops, use built-in functions and libraries)
- Use type aliases to give semantic meaning to primitive types (e.g., `ModelPath = str`)
- Functions should typically return a singular type rather than unions
- Functions not used outside their module should be prefixed with `_`
- Use context managers (`with` statements) for resource management
- Consider using Protocol classes for structural typing
- Use assert to help type checking when function logic limits return types
- Use typing.cast as a no-op to inform the type checker of a type when needed
- Use schemas instead of args and kwargs 😭 (and to replace Anys)


Clue88:
Screenshot 2024-11-07 at 3 33 19 PM

backend/gsr_booking/views.py
```diff
@@ -37,83 +49,87 @@ class UserViewSet(viewsets.ReadOnlyModelViewSet):
...
         return Response({"message": "invite(s) sent."})
 
     @action(detail=True, methods=["post"])
-    def accept(self, request, pk=None):
+    def accept(
+        self, request: Request, pk: Optional[int] = None
+    ) -> Response | HttpResponseForbidden:
         membership = get_object_or_404(GroupMembership, pk=pk, accepted=False)
-        if membership.user is None or membership.user != request.user:
+        user = cast("UserType", request.user)
```
vcai122, line 131: These casts (here and below) are inconsistent. It also seems clunky that are forced to use `cast`. Is this common? If not maybe there's a better way to deal with our User type?

backend/gsr_booking/api_wrapper.py
```diff
@@ -2,24 +2,31 @@
...
 from requests.exceptions import ConnectionError, ConnectTimeout, ReadTimeout
 
-from gsr_booking.models import GSR, GroupMembership, GSRBooking, Reservation
+from gsr_booking.models import GSR, Group, GroupMembership, GSRBooking, Reservation
 from gsr_booking.serializers import GSRBookingSerializer, GSRSerializer
 from utils.errors import APIError
 
 
-User = get_user_model()
+if TYPE_CHECKING:
```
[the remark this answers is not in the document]
vcai122: I'm going to guess its because Django makes us do `get_user_model()`? 

I feel like its clunky to have to put this at the top of all the files.. Maybe we can put this in a file and just import it everywhere? I'm starting to wonder how other Django projects do this cleanly

backend/gsr_booking/views.py
```diff
@@ -190,11 +210,12 @@ class CheckWharton(APIView):
 
     permission_classes = [IsAuthenticated]
 
-    def get(self, request):
+    def get(self, request: Request) -> Response:
+        user: Any = request.user
```
vcai122, line 214: why is this one not casted?

backend/penndata/serializers.py
```diff
@@ -10,6 +12,9 @@
 )
 
 
+ValidatedData: TypeAlias = dict[str, Any]
```
vcai122, line 15: I feel like we don't want this type alias here (also I don't think you did above). Instead, if we are going to have this alias, we should have it in some common shared type file (same with the weird User type thing we are doing). 

But I think again, I'm wondering if we should be doing this "ad-hoc"-ly ourselves, and what existing large production type-safe Django projects do?

backend/penndata/views.py
```diff
@@ -28,12 +31,18 @@
 )
 
 
+ValidatedData: TypeAlias = dict[str, Any]
+CalendarEventList: TypeAlias = QuerySet[CalendarEvent, Manager[CalendarEvent]]
+EventList: TypeAlias = QuerySet[Event, Manager[Event]]
+HomePageOrderList: TypeAlias = QuerySet[HomePageOrder, Manager[HomePageOrder]]
```
vcai122, line 37: Same with these. I know they are for the sake of clarity, but I feel like its a bit overkill so just adding clutter here. And also in Django world, seeing that it is a queryset is actually probably more helpful.

backend/pennmobile/admin.py
```diff
@@ -1,11 +1,20 @@
...
+from django.db.models import Model
+from django.http import HttpRequest
 from django.urls import reverse
 from django.utils.html import format_html
 
 
-def add_post_poll_message(request, model):
+ModelType: TypeAlias = Type[Model]
+AdminContext: TypeAlias = Dict[str, Any]
+MessageText: TypeAlias = str
```
vcai122, line 14: Same here, ifl ModelType should be in some common library (or better, installed library someone else has already put together). AdminContext would also benefit from coming from a common library (maybe more generically as `Context` as this is a common Django idea, or again better from some outside library). MessageText seems overkill.

backend/sublet/serializers.py
```diff
@@ -162,13 +169,19 @@ class Meta:
             "images",
         ]
```
vcai122, line 171: Is there any chance serializers are not returning dictionaries?

backend/sublet/views.py
```diff
@@ -292,15 +301,16 @@ def create(self, request, *args, **kwargs):
 
         return Response(serializer.data, status=status.HTTP_201_CREATED)
 
-    def destroy(self, request, *args, **kwargs):
+    def destroy(self, request: Request, *args: Any, **kwargs: Any) -> Response:
         queryset = self.get_queryset()
         filter = {"user": self.request.user.id, "sublet": int(self.kwargs["sublet_id"])}
-        obj = get_object_or_404(queryset, **filter)
+        obj: Offer = get_object_or_404(queryset, **filter)
```
vcai122, line 307: What dictates if you explicitly annotate the type or not?

vcai122 (commented):
Thanks for doing this!!! I have not worked with type systems in python before so a lot of it is just questions and my initial thoughts. PLEASE feel free to tell me I'm completely wrong since I assume you've been working with this stuff a lot :). 

I am pretty opinionated about code though and I think its really important that we have a really clear story of how we want to use types, so here's my thoughts so far 

1. Obviously types are an overall plus! Especially I really liked the parts that I saw where we had mostly **pure python**: api_wrappers, management commands, helper functions (types on `get_usage` help a lot)
2. I am scared of the `Any`s. I think ideally code with `Any`s should be rewritten/thought about so we can clean these up, but if we just ship it with `Any`s, we remove an incentive to fix it. BUT
   - a lot are on dicts where its a Django dict which seems better (and we can fix it if we use some sort Django type library???)
   - Tuneer suggested we can just have a rule where moving forward ppl are not allowed `Any`s and any code changes touching `Any` code must be fixed.
   - A lot of it is on pretty bad code that needs to be rewritten anyway
3. I am concerned about being too "ad-hoc" about this. I'm talking about the Django part of the code here. I think if we want types to actually be helpful, and not just extra developer work without the benefits, we should be deliberate about the way we are adding types . Because Django comes with so much pre-built stuff and its so opinionated, we should consult what other people have done. To that end I've done some initial research:
     - https://github.com/typeddjango/djangorestframework-stubs/tree/master/rest_framework-stubs seems promising but idk how to use it
     - I think we need some examples of big production Django codebases that uses type safety we can model ours off of
     - If we have to write our own types, we should be deliberate and put it in its own library

Let me know what you think

backend/sublet/views.py
ashleyzhang01: If the type of a variable cant be easily inferred (like x=5 is obv an int) or is ambiguous (None or type can change), it might require an explicit annotation. Or some packages we use may not use type annotations, so when we call those functions, we may need to specify. Or, sometimes types may be ambiguous, like ints and floats are often convertible or strs and datetimes if in right format, so you might want to declare type to increase guardrailing

backend/gsr_booking/api_wrapper.py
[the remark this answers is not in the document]
ashleyzhang01: it's cause get_user_model() is a function, so it gets annoyed when we try to declare its type. but yes put it in another file under utils/types

backend/sublet/serializers.py
ashleyzhang01: if this is about the `to_representation` thing, it seems to be built into rest_framework to convert

backend/gsr_booking/admin.py
```diff
@@ -9,13 +14,14 @@ class GroupMembershipInline(admin.TabularInline):
 
     readonly_fields = ["name"]
 
-    def name(self, obj):
-        return obj.user.get_full_name()
+    def name(self, obj: GroupMembership) -> str:
+        user = cast(UserType, obj.user)
```
tuneerroy, line 18: does this force cast the object? what is the difference between this and `user: UserType = obj.user`?

backend/gsr_booking/management/commands/get_reservations.py
```diff
@@ -83,7 +85,7 @@ def handle(self, *args, **kwargs):
 
         self.stdout.write(f"Number of reservations: {reservations.count()}")
 
-    def __convert_date(self, date_str):
+    def __convert_date(self, date_str: str) -> datetime | None:
```
tuneerroy, line 88: thoughts on using `Optional[datetime]` instead?

tuneerroy (commented):
In general, just concerned with using `Any` types in so many places. feels a bit like spamming `any` in TypeScript to the point that it's basically JavaScript. also need to re-merge main into this branch

dr-Jess:
This is simply a really scary pr to merge. But I like the idea--maybe will look into chunking by app into separate PRs

PR #318, opened 2024-11-06
Minghan market two routes

minghansun1:
Made two routes for Item/Sublet, also made sure other routes work


backend/market/views.py
```diff
@@ -1,37 +1,48 @@
...
 from pennmobile.analytics import Metric, record_analytics
 
 
 User = get_user_model()
 
 
-class Amenities(generics.ListAPIView):
-    serializer_class = AmenitySerializer
-    queryset = Amenity.objects.all()
+class Tags(generics.ListAPIView):
```
vcai122, line 33: hm for these (Tags and Categories) lets just make this a normal APIView with 1 get request chuck out the queryset and serializer_class, and just do like `return Tags.objects.all().values_list(flat=True)`

backend/market/views.py
```diff
@@ -57,167 +68,163 @@ def get_queryset(self):
...
+        item = get_object_or_404(Item, id=int(self.kwargs["item_id"]))
+        return ItemImage.objects.filter(item=item)
 
     # takes an image multipart form data and creates a new image object
     def post(self, request, *args, **kwargs):
         images = request.data.getlist("images")
-        sublet_id = int(self.kwargs["sublet_id"])
-        self.get_queryset()  # check if sublet exists
+        item_id = int(self.kwargs["item_id"])
+        self.get_queryset()  # check if item exists
```
vcai122, line 212: uh idk what this self.get_queryset() is doing?

backend/market/views.py
```diff
@@ -57,167 +68,163 @@ def get_queryset(self):
...
         img_serializers = []
```
vcai122, line 213: change this to list comprehension

backend/market/views.py
```diff
@@ -57,167 +68,163 @@ def get_queryset(self):
...
+            tag_field="item__tags__name",
+        )
 
         # Serialize and return the queryset
-        serializer = SubletSerializerSimple(queryset, many=True)
+        serializer = self.get_serializer(queryset, many=True)
         return Response(serializer.data)
 
 
 class CreateImages(generics.CreateAPIView):
```
vcai122, line 195: Looking at this code we are not really leveraging the fact that it is a CreateAPIView at all... let's pin this and put a TODO about this

backend/market/views.py
```diff
@@ -230,77 +237,75 @@ def destroy(self, request, *args, **kwargs):
 
 
 class Favorites(mixins.DestroyModelMixin, mixins.CreateModelMixin, viewsets.GenericViewSet):
```
vcai122, line 239: This is sus, we are not leveraging the viewset/serializer at all, put another pin here (or fix if you want this fix looks simpler)

backend/market/views.py
```diff
@@ -57,167 +68,163 @@ def get_queryset(self):
...
+    else:
+        queryset = queryset.filter(item__category__name__in=["Sublet"])
+        if params.get("seller", "false").lower() == "true" and user:
+            filters["item__seller"] = user
+        else:
+            filters["item__expires_at__gte"] = timezone.now()
+
+    queryset = queryset.filter(**filters)
+
+    # Apply tag filtering iteratively if "tags" parameter is provided
```
vcai122, line 100: why r tags special

backend/market/views.py
```diff
@@ -57,167 +68,163 @@ def get_queryset(self):
         return Offer.objects.filter(user=user)
 
 
-class Properties(viewsets.ModelViewSet):
+def apply_filters(
```
vcai122, line 71: I think passing this should be called create_filters and we shouldn't be passing in a queryset.
Also filter_mappings, is_sublet, and tag_field is a bit contrived. 

You should 
1. Make a (static) function that does this for Items and does it cleanly and well without extra accessories (in the Item view)
2. Make a function that does this cleanly and well for sublet (in the sublet view) that uses the Item function for help to make it more concise

vcai122 (changes requested):
Okay, code is looking a lot cleaner, really great work!

I left a decent amount of comments all just pushing you towards cleaner code and to understand what Django is actually doing, it's tough thinking about this stuff but you are doing a great job.

Also, I am mentioning a decent amount of stuff that was just code you moved over. For now since they work and are tested lets just leave a TODO to fix in a later PR (perhaps after you get everything in and working)

backend/market/serializers.py
```diff
@@ -1,16 +1,27 @@
...
     class Meta:
-        model = Amenity
+        model = Tag
         fields = "__all__"
 
 
+class CategorySerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Category
+        fields = ["name"]
```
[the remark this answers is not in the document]
minghansun1: The example in django docs included `'account_name'` in both `fields` and `read_only_fields` so I did the same. I'll keep both fields for now, but this isn't required tho so I can remove `"name"` from fields if you prefer that implementation.
https://www.django-rest-framework.org/api-guide/serializers/#specifying-read-only-fields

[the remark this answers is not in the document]
minghansun1: Actually tried to I remove `name` from fields and only declare a read-only field, but apparently declaring a ModelSerializer without a `fields` attribute is deprecated

backend/market/serializers.py
```diff
@@ -50,46 +61,41 @@ def get_image_url(self, obj):
...
+class ItemSerializer(serializers.ModelSerializer):
+    # amenities = ItemSerializer(many=True, required=False)
+    images = ItemImageURLSerializer(many=True, required=False)
+
+    tags = serializers.PrimaryKeyRelatedField(many=True, queryset=Tag.objects.all(), required=False)
+    category = serializers.PrimaryKeyRelatedField(queryset=Category.objects.all(), required=True)
+    seller = serializers.PrimaryKeyRelatedField(queryset=User.objects.all(), required=False)
+    favorites = serializers.PrimaryKeyRelatedField(
+        many=True, queryset=User.objects.all(), required=False
     )
```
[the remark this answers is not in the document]
minghansun1: Yes removing those fields doesn't change functionality in this case. The Item model determines whether a field is required or not.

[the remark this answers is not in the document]
minghansun1: Just to clarify my understanding, we don't want to return seller info on GET requests since it's not relevant until a buyer makes an offer (which should be differentiated from a general GET request).
Is this correct?

backend/market/serializers.py
```diff
@@ -50,46 +61,41 @@ def get_image_url(self, obj):
...
-        model = Sublet
-        read_only_fields = [
-            "id",
-            "created_at",
-            "subletter",
-            "sublettees",
-            # "images"
-        ]
+        model = Item
+        read_only_fields = ["id", "created_at", "seller", "buyer", "images"]
```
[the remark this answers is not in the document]
minghansun1: Buyers is initialized so that it's empty upon creation of an object. It's a `ManyToMany` field, so if it isn't passed to `ModelSerializer's` create method, `ModelSerializer` will just not update the association table, which implicitly initializes buyers to an empty set. You could double check this, I read about this in `ModelSerializer's` source code at line 1011.

Seller is manually set to the user which made the request before being added to validated_data and passed to `ModelSerializer's` create function. You can think of`read_only_field` as being read-only from the user's point of view (it won't be updated based on user input).

BTW i realized that `read_only_fields` should contain `buyers`, not `buyer`

backend/market/serializers.py
```diff
@@ -106,84 +112,121 @@ def contains_profanity(self, text):
         return predict([text])[0]
 
     def create(self, validated_data):
-        validated_data["subletter"] = self.context["request"].user
+        validated_data["seller"] = self.context["request"].user
+        category_name = validated_data.pop("category")
```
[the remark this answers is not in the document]
minghansun1: Huh, I deleted the section where I explicitly initialized Category and my code still seems to work. When I was writing this code, I remember some error saying that Category wasn't expecting a Category object, not a string so I asked GPT to fix it.

I'll try to figure out how initializing another instance of a model works under the hood.

backend/market/serializers.py
```diff
@@ -106,84 +112,121 @@ def contains_profanity(self, text):
...
+        category_instance = Category.objects.get(name=category_name)
+        validated_data["category"] = category_instance
         instance = super().create(validated_data)
         instance.save()
         return instance
 
     # delete_images is a list of image ids to delete
     def update(self, instance, validated_data):
```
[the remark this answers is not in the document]
minghansun1: There's no permission checking done here, it's just manually altering validated_data before sending it to ModelSerializer create.

backend/market/serializers.py
```diff
@@ -106,84 +112,121 @@ def contains_profanity(self, text):
...
             or self.context["request"].user.is_superuser
         ):
             instance = super().update(instance, validated_data)
             instance.save()
             return instance
         else:
-            raise serializers.ValidationError("You do not have permission to update this sublet.")
+            raise serializers.ValidationError("You do not have permission to update this item.")
 
     def destroy(self, instance):
```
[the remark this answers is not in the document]
minghansun1: Yeah I'm not sure why permission checking is done in serializers. Django docs say that permission checks usually run in views

backend/market/serializers.py
```diff
@@ -106,84 +112,121 @@ def contains_profanity(self, text):
...
-    amenities = serializers.PrimaryKeyRelatedField(
-        many=True, queryset=Amenity.objects.all(), required=False
-    )
-    images = SubletImageURLSerializer(many=True, required=False)
+# simple tag serializer for use when pulling all serializers/etc
+class ItemSerializerSimple(serializers.ModelSerializer):
+    tags = serializers.PrimaryKeyRelatedField(many=True, queryset=Tag.objects.all(), required=False)
+    images = ItemImageURLSerializer(many=True, required=False)
+    category = serializers.PrimaryKeyRelatedField(queryset=Category.objects.all(), required=True)
+    seller = serializers.PrimaryKeyRelatedField(queryset=User.objects.all(), required=False)
```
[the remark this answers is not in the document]
minghansun1: Doesn't seem like it, I removed them all

backend/market/serializers.py
```diff
@@ -106,84 +112,121 @@ def contains_profanity(self, text):
...
+                if tags is not None:
+                    instance.item.tags.set(tags)  # Use .set() for many-to-many fields
+
+            # Update the remaining Sublet fields
+            for attr, value in validated_data.items():
+                setattr(instance, attr, value)
+            instance.save()
+            return instance
+
+    def destroy(self, instance):
```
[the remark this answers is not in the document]
minghansun1: Didn't we want to separate sublet operations from item operations? Wouldn't using the destroy item route to destroy sublet make things inconsistent?

backend/market/serializers.py
```diff
@@ -106,84 +112,121 @@ def contains_profanity(self, text):
...
+
+class SubletSerializer(serializers.ModelSerializer):
+    item = ItemSerializer(required=True)
+
+    class Meta:
+        model = Sublet
+        fields = ["id", "item", "address", "beds", "baths", "start_date", "end_date"]
+        read_only_fields = ["id"]
+
+    def create(self, validated_data):
```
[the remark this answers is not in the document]
minghansun1: Django doesn't support writable nested fields, so we need to explicitly state how to create Item.
However, I can call super.create to create the sublet object instead of creating it myself.

backend/market/views.py
minghansun1: It calls the custom get_queryset defined above. If there is no Item with the inputted ID, self.get_queryset will return a `404 Not Found`.

backend/market/views.py
minghansun1: Is this because the whole point of using CreateAPIView is to use the generic create method that Django provides?

backend/market/views.py
minghansun1: They all have the same field name (tags__name for items, item__tags__name for sublets)

backend/market/serializers.py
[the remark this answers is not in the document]
vcai122: gotcha, this looks good then!

backend/market/serializers.py
```diff
@@ -106,84 +102,100 @@ def contains_profanity(self, text):
...
+        item_serializer = ItemSerializer(data=validated_data.pop("item"), context=self.context)
+        item_serializer.is_valid(raise_exception=True)
+        validated_data["item"] = item_serializer.save()
+        instance = super().create(validated_data)
+        return instance
+
+    def update(self, instance, validated_data):
```
vcai122, line 165: I think we can use the item serializer for update just like you did above? 

Also if you find that we don't need this explicit permission check on the serializer, lets remove it

backend/market/serializers.py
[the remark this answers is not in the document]
vcai122: a little bit. Doesn't seem terrible because our end user is in the same room as us (ios/android developers), so we have the liberty of assuming they aren't clueless, and we would be removing room for error in our code. 

but i could go either way. I did see you overode delete on models which is good if we want to go the delete sublet route as that would ensure deletion on admin portal will also delete the item. So you have no need to double delete as you are here. Also clean up this method lol (and remove permission checking if you verify we don't need)

vcai122 (changes requested):
One last round I think. Mainly has to due with verifying if we need the manual permissions check. Your code is look really good, I can see the effort you put in to make it clean. Awesome job!

backend/market/serializers.py
```diff
@@ -50,47 +62,31 @@ def get_image_url(self, obj):
...
-            "start_date",
-            "end_date",
             "expires_at",
-            # "images",
-            # images are now created/deleted through a separate endpoint (see urls.py)
-            # this serializer isn't used for getting,
-            # but gets on sublets will include ids/urls for images
+            "images",
         ]
+        read_only_fields = ["id", "created_at", "seller", "buyers", "images"]
```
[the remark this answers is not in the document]
minghansun1: I added favorites to read-only fields since it should only be changed through the favorites route.

backend/market/serializers.py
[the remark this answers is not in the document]
minghansun1: Oh yeah sorry i thought you were talking about the create function, mb.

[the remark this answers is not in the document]
minghansun1: The update and delete methods are not required here, I deleted both of them.

backend/market/serializers.py
minghansun1: Fixed

backend/market/permissions.py
```diff
@@ -13,32 +13,47 @@ def has_permission(self, request, view):
...
 
     def has_object_permission(self, request, view, obj):
-        # Check if the user is the owner of the Sublet.
-        return request.method in permissions.SAFE_METHODS or obj.sublet.subletter == request.user
+        # Check if the user is the owner of the Item.
+        return (
+            request.method in permissions.SAFE_METHODS
+            or (hasattr(obj, "seller") and obj.seller == request.user)
+            or (hasattr(obj, "item") and obj.item.seller == request.user)
+        )
```
vcai122, line 56: Why would an image have a seller?

backend/market/serializers.py
```diff
@@ -106,84 +100,111 @@ def contains_profanity(self, text):
...
 
 
-class SubletSerializerRead(serializers.ModelSerializer):
-    amenities = serializers.PrimaryKeyRelatedField(
-        many=True, queryset=Amenity.objects.all(), required=False
-    )
-    images = SubletImageURLSerializer(many=True, required=False)
+# Read-only serializer for use when reading a single item
+class ItemSerializerPublic(serializers.ModelSerializer):
+    buyer_count = serializers.SerializerMethodField()
```
vcai122, line 110: why show buyer_count/favorite_count? Do we want this to be public information? Also better name here is offer_count. 

I think we probably don't need to show this, especially if this is the "Public" serializer.

backend/market/serializers.py
```diff
@@ -106,84 +100,111 @@ def contains_profanity(self, text):
...
 
 
-# simple sublet serializer for use when pulling all serializers/etc
-class SubletSerializerSimple(serializers.ModelSerializer):
-    amenities = serializers.PrimaryKeyRelatedField(
-        many=True, queryset=Amenity.objects.all(), required=False
-    )
-    images = SubletImageURLSerializer(many=True, required=False)
+# Read-only serializer for use when pulling all items/etc
+class ItemSerializerList(serializers.ModelSerializer):
```
vcai122, line 140: If this is read only why not make all the fields read only.

backend/market/serializers.py
vcai122: What is this public serializer for. Why is it different than the list serializer?

backend/tests/market/test_market.py
```diff
@@ -0,0 +1,1283 @@
...
+    def setUp(self):
+        self.user = User.objects.create_user("user", "user@seas.upenn.edu", "user")
+        self.client = APIClient()
+        self.client.force_authenticate(self.user)
+        user1 = User.objects.create_user("user1", "user1@seas.upenn.edu", "user1")
+        user2 = User.objects.create_user("user2", "user2@seas.upenn.edu", "user2")
+        tags = ["New", "Used", "Couch", "Laptop", "Textbook", "Chair", "Apartment", "House"]
+        categories = ["Book", "Electronics", "Furniture", "Food", "Sublet", "Other"]
+        for tag in tags:
+            Tag.objects.create(name=tag)
```
vcai122, line 30: use bulk create and list comprehension

backend/tests/market/test_market.py
```diff
@@ -0,0 +1,1283 @@
...
+        for category in categories:
+            Category.objects.create(name=category)
+        # "backend/tests/market/mock_items.json" if debugging
+        # "tests/market/mock_items.json" if from backend directory
+        # item 1, 5 are self.user. item 3, 4 are user1. item 2, 6 are user2
+        with open("tests/market/mock_items.json") as data:
```
vcai122, line 36: i think its better to create directly, i dont see a strong reason to use these files

vcai122: then you can also avoid the weird enumerate and checking the index

vcai122 (changes requested):

backend/market/permissions.py
minghansun1: The reason I did that was because I incorrectly set the permission class of the `CreateImages` view to`ItemImageOwnerPermission`, when it should have been `ItemOwnerPermission`, since `ItemImage` hasn't been initalized yet.
I changed the permission class of `CreateImages`, and fixed this code

backend/market/serializers.py
minghansun1: I think that most ecommerce website have `buyer_count`+`favorite_count` to show how popular an `Item` is. Not sure if it's as applicable in our case since most users aren't trying to sell at scale, but it's probably a good feature to test out.

minghansun1: `ListSerializer` only contains the most important information about an Item which a user would pay attention to while scrolling a catalog of items, whereas `PublicSerializer` contains the full list of publicly accessible info about an item.
Frontend would use ListSerializer for a page listing all items, and PublicSerializer for a more detailed product view page.

Alternatively, we can get rid of PublicSerializer, add a feature to filter by item_id/sublet_id to the ListFeature, and add all important fields to ListSerializer. Would eliminate a serializer, but might cause confusion/result in large amounts of data being passed over network

minghansun1: If we wanted, we could rewrite all 3 Serializers so that we pass in a flag which specifies whether it's a public, private, or list serializer and adjust the fields during Serializer initialization.
image

backend/market/serializers.py
minghansun1: Pretty sure all fields are read-only?
image

backend/tests/market/test_market.py
minghansun1: > i think its better to create directly, i dont see a strong reason to use these files

I think this adds modularity since the testing items are stored separately. Also makes the code in test_market shorter a lot shorter and easier to read

vcai122 (approved):
LETS GOO

PR #323, opened 2024-11-11
Remove Me Groups

vcai122:
We did a GSR refactor that removed the need for "Me" groups last year
- "Me" groups existed in the first place so we could treat booking GSRs for users the same as booking GSRs for groups.
- Turns out the logic can be optimized if we separate the use cases so we did just that in the GSR refactor, using `group=None` to just book the group for the given user. 

This PR removes the the leftovers to __erase the idea of "Me" groups completely from our codebase__. 

Also did a small refactor with a group membership view. 
- __Note__: I checked iOS code and `users` (now `/mymemberships`), `membership`, and `groups` isn't used at all
  - We probably will never implement group feature. But its probably worth refactoring this code so that it makes sense, _eventually_. Since if we are going to keep it, might as well avoid confusion and make the business logic clear. Right now it is pretty messy.

### DIFF
--- a/backend/gsr_booking/serializers.py
+++ b/backend/gsr_booking/serializers.py
@@ -20,21 +20,13 @@ class GroupRoomBookingRequestSerializer(serializers.Serializer):
         return obj["lid"] == 1
 
 
-class MiniUserSerializer(serializers.ModelSerializer):
-    class Meta:
-        model = User
-        fields = ["username", "first_name", "last_name"]
-
-
 class GroupMembershipSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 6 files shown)


tuneerroy (approved):
LGTM! question: if we remove groups, how are we planning on supporting the Labs shared group?

vcai122:
> LGTM! question: if we remove groups, how are we planning on supporting the Labs shared group?

We aren't removing Groups. Groups are still alive and well. We are just removing the idea of "Me" groups as a singleton group, because we want to handle the logic of a single user booking with their own credits separately from group logic.

PR #338, opened 2025-02-16
uv migration

clay53:


dr-Jess (commented):
I want to approve this but I'm scared it will bork our dependencies/lots of things. How should I verify?

PR #341, opened 2025-02-27
Added analytics to booking and laundry APIs and removed previous analytics tracking functionalities

aangelinali:
Added analytics to booking and laundry APIs to record GSR booking durations and locations, added analytics for laundry washing time minutes by dorm, removed previous analytics tracking functionalities


backend/gsr_booking/views.py
```diff
@@ -195,11 +199,33 @@ def get(self, request, lid, gid):
...
+            ),
+            get_value=lambda args, res: (
+                (
+                    parse_datetime(args[1].data["end_time"])
+                    - parse_datetime(args[1].data["start_time"])
+                ).total_seconds()
+                / 60
+            ),
+        )
+    )
```
dr-Jess, line 228: The logic here doesn't seem to be triggering for testing--needs to be fixed

backend/laundry/views.py
```diff
@@ -83,6 +85,24 @@ def get_snapshot_info(room_id):
         snapshots = LaundrySnapshot.objects.filter(filter).order_by("-date")
         return (room, snapshots)
```
dr-Jess, line 87: Let's come up with a more consistent naming schema. Potentially subproject (so gsr_booking, laundry) as head, so like `<subproject>.<details>.<type (ex. duration)>`

dr-Jess (changes requested):
Also needs to be unit tested after changes are made. Good work so far!

backend/laundry/views.py
vcai122: No need for naming scheme. Library handles it intentionally so developers do not have to waste time thinking about scheme

vcai122 (commented):
i looked for like 10 seconds ill look more later

backend/laundry/api_wrapper.py
```diff
@@ -123,6 +125,36 @@ def room_status(room):
...
+    )
+
+
+@LabsAnalytics.record_function(
+    FuncEntry(
+        name="cron.full_laundry_room",
+        get_value=lambda args, res: sum(1 for room_data in res.values() if is_room_full(room_data)),
+    ),
+    FuncEntry(
+        name="cron.full_laundry_room_timestamp",
```
dr-Jess, line 144: This isn't necessary--timestamps are already automatically stored, and the previous entry already captures necessary info the SQL query this entry from.

backend/penndata/views.py
```diff
@@ -144,6 +146,10 @@ def get_queryset(self):
         return HomePageOrder.objects.all()
 
 
+# Tracks the time user opens up Penn Mobile homepage
+@LabsAnalytics.record_apiview(
+    ViewEntry(name="homepage_visit_time", get_value=lambda req, res: timezone.localtime())
```
dr-Jess, line 151: ^ same as above, can maybe just store a "1" or something and we can SQL query this info later.

dr-Jess (changes requested):

backend/laundry/api_wrapper.py
aangelinali: Got it, I've removed them so it only stores 1 every time the endpoint is hit

dr-Jess (approved):
lgtm!! great work

PR #342, opened 2025-02-28
Added health route to backend

sritanmotati:
Added a health route to the backend
* `/api/health`
* Added test case


vcai122:
> Added a health route to the backend
> 
> * `/api/health`
> * Added test case

Why are we adding to each product separately instead of DLA

PR #343, opened 2025-03-02
fixed offer permissions

minghansun1:


vcai122 (approved):
LGTM

backend/market/views.py
```diff
@@ -299,5 +308,6 @@ def destroy(self, request, *args, **kwargs):
     def list(self, request, *args, **kwargs):
         if not Item.objects.filter(pk=int(self.kwargs["item_id"])).exists():
             raise exceptions.NotFound("No Item matches the given query")
-        self.check_object_permissions(request, Item.objects.get(pk=int(self.kwargs["item_id"])))
+        for offer in self.get_queryset():
+            self.check_object_permissions(request, offer)
```
vcai122, line 312: i dont think you need to check manually... won't setting the permission class already check?

backend/market/views.py
```diff
@@ -299,5 +308,6 @@ def destroy(self, request, *args, **kwargs):
     def list(self, request, *args, **kwargs):
         if not Item.objects.filter(pk=int(self.kwargs["item_id"])).exists():
             raise exceptions.NotFound("No Item matches the given query")
-        self.check_object_permissions(request, Item.objects.get(pk=int(self.kwargs["item_id"])))
+        for offer in self.get_queryset():
```
vcai122, line 311: On another note, do you have to manually raise NotFound exception? If you do, shouldn't you be raising in the queryset?

vcai122, line 311: I feel like we don't need to manually override list if we do the queryset and object permissions right

backend/market/views.py
```diff
@@ -299,5 +308,6 @@ def destroy(self, request, *args, **kwargs):
     def list(self, request, *args, **kwargs):
```
vcai122, line 308: Looking at the destroy code on line 303, it seems like you are doing some manual checks to see if the object can be destroyed. Again, I think if we get the permissions right (which I think they are) and the queryset right (which I think you have correct as well), Django's built in destroy should be able to handle all of it for us?

PR #349, opened 2025-03-17
Fix CI/CD Workflow

jonathanmelitski:
Due to deprication of GH artifacts-related actions, some changes were needed to allow the images to be published.

**The problem:**
For some reason, the docker build step started creating artifacts related to its own builds. These were not present in previous successful workflow runs (so I'm inclined to believe that Docker recently changed this).
image
The workflow we use only uses the `build-frontend` and `build-backend` images uploaded in the previous steps. Further, the download action fails to download these Docker-uploaded artifacts, leading to a failure for artifacts that are not even used.

**The fix:**
`actions/download-artifact@v4` has a new flag called `pattern` which allows the use of a filter. As such, we can filter for `build-frontend` and `build-backend` by passing in this parameter. This forces only downloading these two artifacts, which will succeed.
image

https://github.com/pennlabs/penn-mobile/actions/runs/13888718422/job/38857154032

### DIFF
--- a/.github/workflows/build-and-deploy.yaml
+++ b/.github/workflows/build-and-deploy.yaml
@@ -92,6 +92,14 @@ jobs:
       steps:
         - uses: actions/checkout@v4
         - uses: actions/download-artifact@v4
+          with:
+            pattern: build-backend
+        - uses: actions/download-artifact@v4
+          with:
+            pattern: build-frontend
+        - uses: actions/download-artifact@v4
+          with:
+            pattern: pennlabs~penn-mobile~*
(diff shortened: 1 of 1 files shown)


jonathanmelitski:
Another note:

The reason we're facing this issue now is because Docker build-push-action just recently implemented the summaries feature (which uploads the docker builds as artifacts). As a result, they provided the following note (last week) in their `README.md` (https://github.com/docker/build-push-action):

image

joyliu-q (approved):
thanks for looking into this!

PR #356, opened 2025-05-01
make ID serialize with page model

jonathanmelitski:
Include the `id` field in the JSON response for wrapped pages.


dr-Jess (approved):
Looks fine. If it doesn't work will roll back

PR #357, opened 2025-05-07
Update semester field to be longer

dr-Jess:


backend/wrapped/migrations/0005_alter_semester_semester.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ("wrapped", "0004_page_id"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="semester",
+            name="semester",
+            field=models.CharField(max_length=16, primary_key=True, serialize=False),
```
jonathanmelitski, line 16: does `serialize=False` mean it won't be included in the API call? 

image

Frontend expects `string` `semester` to be sent with (even if it's oddly redundant) the pages, and would cause issues if it's not found.

backend/wrapped/models.py
```diff
@@ -26,7 +26,7 @@ class GlobalStatKey(StatKey):
 
 
 class Semester(models.Model):
-    semester = models.CharField(max_length=5, primary_key=True, null=False, blank=False)
+    semester = models.CharField(max_length=16, primary_key=True, null=False, blank=False)
```
jonathanmelitski, line 29: I can't imagine we'd want it to be longer than this in the future. I'm good with this if you are.

jonathanmelitski (changes requested):
Can you address the `serialize=False` comment, then it should be fine. I just want to make sure there is a semester field in the overall API response.

backend/wrapped/migrations/0005_alter_semester_semester.py
dr-Jess: This is an internal config for our DB migrations and has no effect on the actual endpoints. [https://docs.djangoproject.com/en/5.2/topics/migrations/#serializing-values](url)

jonathanmelitski: alr yeah, just making sure.

jonathanmelitski (approved):
LGTM (let's get this money)

PR #360, opened 2025-09-06
Fixed test cases + error return status code

minghansun1:


backend/tests/market/mock_sublets.json
dr-Jess, line 1: remove before merge

backend/tests/market/mock_items.json
dr-Jess, line 1: remove before merge

backend/tests/market/test_market.py
```diff
@@ -16,245 +14,279 @@
 
 User = get_user_model()
 
-class TestMarket(TestCase):
 
+class BaseMarketTest(TestCase):
     def setUp(self):
```
dr-Jess, line 19: We ought to maximize the amount of setup that is done here as opposed to copying code between each case. Please go through each case and find as much shared/stuff you can put here as possible :)

dr-Jess (changes requested):
1-2 small changes off being done. Good stuff and thanks for handling this!

dr-Jess (commented):
lgtm

PR #412, opened 2026-03-23
Decrease API request timeout to 5 seconds

jonathanmelitski:
Reduced timeout for API requests from 16 to 5 seconds.

They fixed the issue (or at least brought it back below 16 seconds per request lol). If this leads to a higher volume of sentry errors we will adjust it again.


minghansun1 (approved):
lgtm

PR #413, opened 2026-03-26
Replace DRF built-in OpenAPI schema generation with drf-spectacular

darrentweng:
## Summary
- `/api/openapi/` and `/api/documentation/` were broken due to a version incompatibility: DRF's built-in schema generation calls `get_schema_operation_parameters()` on filter backends, but `django-filter` [deprecated this in v23.2 and removed it in v25.1](https://github.com/carltongibson/django-filter/blob/main/CHANGES.rst)
- As noted in the [DRF docs](https://www.django-rest-framework.org/topics/documenting-your-api/), "REST framework recommends using third-party packages for generating and presenting OpenAPI schemas, as they provide more features and flexibility than the built-in (deprecated) implementation." Replaced with `drf-spectacular`, the recommended library. It targets OpenAPI 3 (vs the older OpenAPI 2/Swagger spec), has `django-filter` support, and is actively maintained.
- Removed the manually maintained `redoc.html` template

## Endpoints
| Endpoint | Before | After |
|---|---|---|
| `/api/openapi/` | Raw schema (broken) | Raw schema (working) |
| `/api/documentation/` | ReDoc UI (broken) | Swagger UI |
| `/api/redoc/` | did not exist | ReDoc UI |

## Tested
- `python manage.py spectacular --validate` generates schema without crashing (pre-existing warnings about undocumented `APIView`s and unregistered `PlatformAuthentication` are unrelated to this change)
- `/api/documentation/` loads Swagger UI
- `/api/redoc/` loads ReDoc UI

## Next steps
- **Schema completeness:** `--validate` reports many views omitted from the schema due to missing `serializer_class` on plain `APIView`s, untyped path parameters, and `PlatformAuthentication` not having a registered `OpenApiAuthenticationExtension`. These are pre-existing gaps. Each warning points to the exact file/line to fix
- **Double-registered routes:** All routes are registered twice (under `/api/` and at the root `""`, see `/backend/pennmobile/urls.py:51`), so the schema and Swagger UI show every endpoint duplicated. Worth investigating whether the root routes (`/gsr/`, `/dining/`, etc.) are actually used or can be removed
- Note: the duplicate root-level routes (`/gsr/`, `/dining/`, etc.) are unreachable in production. Maybe the reverse proxy only forwards requests with a `/api/` prefix to Django? They do resolve locally via runserver (since there's no proxy in that environment?).


minghansun1 (approved):
lgtm

PR #417, opened 2026-04-05
Minghan/make dining cache clearable, remove old cache

minghansun1:


minghansun1:
i lowkey forgot to remove the old caching in the last pr

backend/pennmobile/settings/base.py
darrentweng, line 13: I don't see anything wrong with having this, but out of curiosity, is there any reason that this is necessary? Dotenv isn't really present in the codebase (except for being a package in the Pipfile.lock), and pipenv shell loads all environment variables.

darrentweng (approved):
LGTM

PR #418, opened 2026-04-05
Grace delete duplicate menus

gracech1:
All duplicate menus are deleted in the load_menus() function. This means that whenever the load_next_menu.py command is called there are no duplicate menus.


gracech1:
Merging between caching changes and my changes are messed up...

PR #420, opened 2026-04-06
Fix/openapi drf spectacular v2

darrentweng:
## Summary

Re-applies the changes from #413, which were reverted in #414 after a `ModuleNotFoundError: cachetools` took down prod.

### Root cause of the #413 outage
`pipenv install drf-spectacular` re-resolved the full dependency graph and silently dropped `cachetools` from `Pipfile.lock`. `google-auth 2.41.1` hard-imports `cachetools` at startup (`jwt.py:53: import cachetools`), so a clean Docker build (which installs only what's in the lock file) crashed immediately on boot.

### What's different this time
This re-resolution also upgraded `firebase-admin` (7.1.0→7.3.0), which transitively pulled `google-auth` up to 2.49.1. That version no longer uses `cachetools` (removed from source, not just metadata) and made `rsa` an optional extra backed by `cryptography` instead. Both drops are verified against the package source, not just `Requires-Dist`.

`pytz` was also silently dropped by the resolver despite being a hard runtime dep of `pandas 2.3.3`, which is the same class of bug. Pinned explicitly in `Pipfile` as a guard. `Pipfile.lock` was regenerated from scratch and all changes are accounted for:

| Package | Change | Reason |
|---|---|---|
| `drf-spectacular` | Added | This PR |
| `jsonschema`, `jsonschema-specifications`, `referencing`, `rpds-py` | Added | drf-spectacular transitive deps |
| `python-discovery` | Added | New dep of virtualenv 21.x (via pre-commit) |
| `pytz` | Added to Pipfile | Guard against pipenv resolver silently dropping it |
| `cachetools` | Removed | google-auth 2.49.1 doesn't import it |
| `rsa` | Removed | google-auth 2.49.1 made it optional; uses `cryptography` instead |

---

- `/api/openapi/` and `/api/documentation/` were broken due to a version incompatibility: DRF's built-in schema generation calls `get_schema_operation_parameters()` on filter backends, but `django-filter` [deprecated this in v23.2 and removed it in v25.1](https://github.com/carltongibson/django-filter/blob/main/CHANGES.rst)
- As noted in the [DRF docs](https://www.django-rest-framework.org/topics/documenting-your-api/), "REST framework recommends using third-party packages for generating and presenting OpenAPI schemas, as they provide more features and flexibility than the built-in (deprecated) implementation." Replaced with `drf-spectacular`, the recommended library. It targets OpenAPI 3 (vs the older OpenAPI 2/Swagger spec), has `django-filter` support, and is actively maintained.
- Removed the manually maintained `redoc.html` template
- Suppressed drf-spectacular schema warnings in production (`DISABLE_ERRORS_AND_WARNINGS: True` in `SPECTACULAR_SETTINGS`) to avoid flooding container stderr with pre-existing warnings about undocumented views and unregistered authenticators

## Endpoints
| Endpoint | Before | After |
|---|---|---|
| `/api/openapi/` | Raw schema (broken) | Raw schema (working) |
| `/api/documentation/` | ReDoc UI (broken) | Swagger UI |
| `/api/redoc/` | did not exist | ReDoc UI |

## Tested
- `pipenv lock` regenerated from scratch, `pipenv sync` applied cleanly
- `python manage.py spectacular --validate` exits 0 (pre-existing warnings about undocumented `APIView`s and unregistered `PlatformAuthentication` are unrelated to this change)
- `/api/documentation/` loads Swagger UI
- `/api/redoc/` loads ReDoc UI

## Next steps
- **Schema completeness:** `--validate` reports many views omitted from the schema due to missing `serializer_class` on plain `APIView`s, untyped path parameters, and `PlatformAuthentication` not having a registered `OpenApiAuthenticationExtension`. These are pre-existing gaps. Each warning points to the exact file/line to fix
- **Double-registered routes:** All routes are registered twice (under `/api/` and at the root `""`, see `backend/pennmobile/urls.py:56`), so the schema and Swagger UI show every endpoint duplicated. Worth investigating whether the root routes are actually used or can be removed


minghansun1 (approved):
LGTM

PR #424, opened 2026-04-26
Grace fix auth token

gracech1:
Changes:
1. To improve auth token refresh, this PR adds a `threading.Lock()` to `DiningAPIWrapper.update_token()` to prevent race conditions where multiple threads attempt to refresh the dining API token simultaneously. It also adds a `TOKEN_EXPIRATION_BUFFER` to prevent the edge case where a token is just _almost_ expired. Test cases were also added for the token refresh logic. 
2. Now we only use one `DiningAPIWrapper` for loading menus! `load_next_menu`  used to make 7 different `DiningAPIWrapper`s (for loading 7 days of dining) but now it only uses one and passes it to the 7 function calls it has.
3. This PR also adds some functionality for dietary restriction search. It adds 20 boolean fields (ex. `vegan`, `halal`, `peanut`) that Penn Dining tracks to the `DiningItem `database model for future use in search. While we were at it, I refactored the `load_items `method to use `_build_dining_item` (for easier loading of allergies and because it was just in a big list comprehension before). This function also uses a `DINING_ICON_IDS `dictionary to automatically map the `cor_icon `IDs from the Penn Dining API to these new fields when building `DiningItem`s. These fields were added to the test cases too. 

Sorry for all the changes on the same branch I won't do it again smhsmh


backend/dining/management/commands/load_next_menu.py
```diff
@@ -32,6 +31,7 @@ def handle(self, *args, **kwargs):
         Load menu for the next 7 days
         """
         today = timezone.now().date()
+        d = DiningAPIWrapper()
```
Divak2004: nit: better variable name, like `dining_api_wrapper`

backend/dining/management/commands/load_next_menu.py
```diff
@@ -32,6 +31,7 @@ def handle(self, *args, **kwargs):
...
 
         for i in range(7):
```
Divak2004: nit: better variable name, like `days_in_advance`. this is existing tho so up to you

backend/dining/api_wrapper.py
```diff
@@ -19,10 +20,38 @@
 OPEN_DATA_URL = "https://3scale-public-prod-open-data.apps.k8s.upenn.edu/api/v1/dining/"
 OPEN_DATA_ENDPOINTS = {"VENUES": OPEN_DATA_URL + "venues", "MENUS": OPEN_DATA_URL + "menus"}
 
+# Dining icon ids for parsing the API response cor_icon field into allergen boolean fields
+DINING_ICON_IDS = {
```
Divak2004, line 24: maybe order this alphabetically so it's consistent with the migration? applies to other files too. up to you tho

backend/dining/api_wrapper.py
```diff
@@ -206,23 +237,31 @@ def load_stations(self, station_response, dining_menu):
             station.items.add(*items)
             station.save()
 
+    def _build_dining_item(self, key, value):
```
Divak2004: nit: better variable naming. maybe you can just change `key` to `item_id` and `value` to `item_info`? i'm not too sure what `value` is. similarly apply this to the loop that calls this method

backend/dining/api_wrapper.py
```diff
@@ -206,23 +237,31 @@ def load_stations(self, station_response, dining_menu):
...
+            item_id=key,
+            name=value["label"],
+            description=value["description"],
+            ingredients=value["ingredients"],
+            allergens=", ".join(value["cor_icon"].values()) if value["cor_icon"] else "",
+            **icon_flags,
+            nutrition_info=json.dumps(
+                {
+                    x["label"]: f"{x['value']}{x['unit']}"
+                    for x in value["nutrition_details"].values()
```
Divak2004: what is `x`? could rename it so it shows what it is

Divak2004 (changes requested):
overall lgtm! i rly like how you refactored a bunch of stuff so the code is cleaner

backend/dining/api_wrapper.py
```diff
@@ -31,16 +60,19 @@ def __init__(self):
     def update_token(self):
         if self.expiration > timezone.localtime():
```
minghansun1: Add a bit of buffer to prevent an edge case where the token's almost expired, but not actually expired when checking the condition, which causes it to return. But the the token could expire before the original request goes through.
It's a very low probability bug but it's better to be safe

backend/dining/api_wrapper.py
gracech1: keeping this just for now because they're roughly listed by categorization (ex. the first few are dining restrictions and the last few are allergies). I may change this is future tho

backend/dining/management/commands/load_next_menu.py
gracech1: done

backend/dining/management/commands/load_next_menu.py
gracech1: done

backend/dining/api_wrapper.py
gracech1: done

backend/dining/api_wrapper.py
gracech1: changed to "detail" because its about nutrition details

minghansun1 (changes requested):

Divak2004 (approved):
lgtm! i don't see anything else but i'm not well versed in mobile backend + i need to f*** off and work on android

backend/dining/api_wrapper.py
gracech1: done. did 5 seconds, let me know if that isn't enough.

gracech1:
> lgtm! i don't see anything else but i'm not well versed in mobile backend + i need to f*** off and work on android

no dont leave its fun here in mobile backend 🚐

minghansun1 (approved):
LGTM

PR #425, opened 2026-09-04
Update AASA to prove pennmobile domain ownership

jonathanmelitski:
Apple has a in-built OAuth handler that requires proof that you own the callback domain using the Apple App Site Association file.

This PR updates that file to include `pennmobile.org`

cc @khoiddinh


minghansun1 (approved):
LGTM

jonathanmelitski:
added reference:

https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession/callback/https(host:path:)

PR #426, opened 2026-09-05
calendar script only gets current semester's events

minghansun1:
Now database doesn't contain spring semester stuff in the fall semester. No need for the nonactive semester's data.

Stack:
[https://github.com/pennlabs/penn-mobile/pull/426](https://github.com/pennlabs/penn-mobile/pull/426)
[https://github.com/pennlabs/penn-mobile/pull/431](https://github.com/pennlabs/penn-mobile/pull/431)

### DIFF
--- a/backend/penndata/management/commands/get_calendar.py
+++ b/backend/penndata/management/commands/get_calendar.py
@@ -70,8 +70,7 @@ class Command(BaseCommand):
                 date = datetime.datetime.strptime(
                     month + day + str(current_year) + "-04:00", "%B%d%Y%z"
                 )
-                if date and date >= timezone.localtime():
-                    CalendarEvent.objects.get_or_create(event=event, date=date_info, date_obj=date)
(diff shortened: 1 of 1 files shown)


jonathanmelitski (dismissed):
This will lead to old events being shown on Penn Mobile frontend. Same comment and thoughts as the ones I made at #427 

In a perfect world (with no versioning). We should only ever add capability to endpoints, since we don't know what downstream frontends depend on the old version.

minghansun1:
> This will lead to old events being shown on Penn Mobile frontend. Same comment and thoughts as the ones I made at #427
> 
> In a perfect world (with no versioning). We should only ever add capability to endpoints, since we don't know what downstream frontends depend on the old version.

No old events will be shown. The `/calendar` route doesn't return all events in the DB, just events with `today` <= `date` <= `today + days_ahead`

darrentweng (approved):

PR #427, opened 2026-09-05
Added days_ahead parameter to calendar route

minghansun1:
## Calendar route: days_ahead param 

The calendar route ([https://pennmobile.org/api/penndata/calendar/](https://pennmobile.org/api/penndata/calendar/)) now serves a list of events for the current semester, sorted by dates.
Use the `days_ahead` query param to adjust how far ahead you want events. No value -> no cutoff

#### Use Case:
Get the semester start and end dates for dining graph, no need to manually update dates!

image
image

Stack:
[https://github.com/pennlabs/penn-mobile/pull/426](https://github.com/pennlabs/penn-mobile/pull/426)
[https://github.com/pennlabs/penn-mobile/pull/427](https://github.com/pennlabs/penn-mobile/pull/427)

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/backend/penndata/views.py
+++ b/backend/penndata/views.py
@@ -102,17 +102,20 @@ class News(APIView):
 
 class Calendar(generics.ListAPIView):
     """
-    list: Returns upcoming university events (within 2 weeks away)
+    list: Returns upcoming university events, configurable cutoff
     """
 
     permission_classes = [AllowAny]
     serializer_class = CalendarEventSerializer
 
     def get_queryset(self):
+        days_ahead = int(
+            self.request.query_params.get("days_ahead", 365)
+        )  # db contains current semester, 1 yr guarantees everything's returned
         return CalendarEvent.objects.filter(
             date_obj__gte=timezone.localtime(),
-            date_obj__lte=timezone.localtime() + timedelta(days=30),
-        )
+            date_obj__lte=timezone.localtime() + timedelta(days=days_ahead),
+        ).order_by("date_obj")
 
 
 class Events(generics.ListAPIView):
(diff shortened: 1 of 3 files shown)


backend/penndata/views.py
```diff
@@ -102,17 +102,20 @@ def get(self, request):
...
-    list: Returns upcoming university events (within 2 weeks away)
+    list: Returns upcoming university events, configurable cutoff
     """
 
     permission_classes = [AllowAny]
     serializer_class = CalendarEventSerializer
 
     def get_queryset(self):
+        days_ahead = int(
+            self.request.query_params.get("days_ahead", 365)
```
jonathanmelitski, line 113: This will lead to a bit of a weird UX in my opinion until an iOS version to include this parameter is merged, released, and propagated.

This is, logically, the correct way to implement this. We should talk about whether we want the default to be backwards-compatible with what is currently in place (default to 30 days, as to not depend on an iOS update)

@khoiddinh

jonathanmelitski (changes requested):
We definitely should talk about this. In the proposed state, that list of events on the Penn Mobile will include every event from the semester, which is not great UX.

If we think that's fine until a change can be made, I guess we can do this? The other option would be to add the parameter on iOS, let it propagate, then push the change that actually uses the parameter.

Generally speaking, when changing endpoints like this (on a mobile product) we have to consider that changes can't be pushed instantaneously and that the old version still has to be viable (even after that version is pushed, since people stay on old versions for a while)

backend/penndata/views.py
minghansun1: Yeah fair, we can implement versioning for this update. It's a good small change to introduce the practice into our codebase

PR #430, opened 2026-09-06
Add leaderboard sort/limit params, replace username with opt-in display name

bliu8:
Leaderboard route now also has sort and limit params so frontend has more optionality on what they want to show. There is also now a tiebreaker for users with the same score.

Players are anonymized by default on the leaderboard, and if they opt in their full names get shown on the leaderboard.


backend/games/models.py
```diff
@@ -27,6 +27,7 @@ class LeaderboardEntry(models.Model):
 
     score = models.PositiveIntegerField(db_index=True)
     num_words_found = models.PositiveIntegerField()
+    show_name = models.BooleanField(default=False)
```
minghansun1, line 30: The decision for whether a name is shown should be tied to the user, not individual submissions (mostly for simplicity). It should be be stored with the user's info.

Consider looking into if we can make a user table just for games (try django inheritance)

backend/games/views.py
```diff
@@ -9,6 +9,9 @@
 from pennmobile.analytics import LabsAnalytics
 
 
+LEADERBOARD_SORT_FIELDS = ("score", "num_words_found", "submitted_at")
```
minghansun1, line 12: Why should the frontend need to sort by time submitted?

backend/games/views.py
```diff
@@ -47,13 +50,29 @@ def get(self, request, date):
 class LeaderboardByDateView(APIView):
     """
     GET: returns the leaderboard for a specific date
+
```
minghansun1, line 53: Any support for filtering by school / other attributes?
It's fine if not, we can easily add it later

backend/games/views.py
```diff
@@ -47,13 +50,29 @@ def get(self, request, date):
...
     def get(self, request, date):
         game = get_object_or_404(Game, date=date)
-        entries = game.scores.all()
+
+        sort = request.query_params.get("sort", "-score")
+        if sort.lstrip("-") not in LEADERBOARD_SORT_FIELDS:
+            return Response(
+                {"error": f"sort must be one of {list(LEADERBOARD_SORT_FIELDS)}."}, status=400
+            )
+        entries = game.scores.select_related("user").order_by(sort, "submitted_at")
```
minghansun1, line 69: You think we should show scores that are anonymous? What if a lot of the top scores are all anonymous?

minghansun1 (commented):
Good work! Happy to see you locked in 😃 
I left a few comments about questions/suggestions I have

backend/games/models.py
```diff
@@ -6,6 +7,73 @@
...
+
+    @classmethod
+    def sync_from_platform(cls, user, show_name=None):
+        attrs = platform_student_attrs(user)
+        if show_name is not None:
+            attrs["show_name"] = show_name
+        return cls.for_user(user, **attrs)
+
+
+class GameUserTag(models.Model):
```
minghansun1, line 65: Why does tag need to be a model? Why not just have a school column and a major column?

IK people might have multiple schools/majors, but we should prob just let them choose 1, or else determining which leaderboard to show might get confusing

Also the code will be much simpler

backend/games/views.py
```diff
@@ -47,14 +93,58 @@ def get(self, request, date):
...
+        year: filter by graduation year
     """
 
     permission_classes = [IsAuthenticated]
 
     def get(self, request, date):
         game = get_object_or_404(Game, date=date)
-        entries = game.scores.all()
-        return Response(LeaderboardEntrySerializer(entries, many=True).data)
+
```
minghansun1, line 109: Nit: This view's code is not very clean, try to see if there are any complexity that can be removed

backend/games/views.py
```diff
@@ -4,11 +4,57 @@
...
 
+LEADERBOARD_SORT_FIELDS = ("score", "num_words_found", "submitted_at")
+
+
+def is_opted_in(user):
+    game_user = getattr(user, "gameuser", None)
+    return bool(game_user and game_user.show_name)
+
+
+def serialize_entry(entry, show_names):
```
minghansun1, line 20: Nit: This function's called once and it's 1 line, does it really need to be a function?

backend/games/views.py
```diff
@@ -72,22 +166,25 @@ def post(self, request, date):
...
-                {"error": "Invalid words submitted.", "invalid_words": invalid}, status=400
+                {"detail": "Invalid words submitted.", "invalid_words": invalid}, status=400
             )
 
         if LeaderboardEntry.objects.filter(game=game, user=request.user).exists():
-            return Response({"error": "Score already submitted for this game."}, status=400)
+            return Response({"detail": "Score already submitted for this game."}, status=400)
+
+        show_name = request.data.get("show_name") is True if "show_name" in request.data else None
+        game_user = GameUser.sync_from_platform(request.user, show_name=show_name)
```
minghansun1, line 187: Consider whether we need to sync from platform for each request. 

The code's getting kinda complex since our route updates a lot of fields that don't necessarily need to be updated.

It might make more sense to just do batch migration to get the school and grad year from platform, and assume that those fields are already populated correctly in the routes.

minghansun1 (commented):
Thanks for the hard work! I have just a few tips

- Have a clear understanding of the responsibilities of each component. For example, SubmitScoreView currently updates user info. We should make this view only responsible for submitting scores. Updating user info can either be another view (if the user wants to manually change their school, or a batch job that syncs from platform on a schedule).
- General code cleanliness, try to minimize unnecessary abstractions and make the code as clear as possible. This is just something that'll improve with practice

PR #433, opened 2026-09-16
fix(is_wharton): return false if 403

esinx:
Fixes is_wharton to return False on 403, also update `APIError` to include status code


backend/utils/errors.py
```diff
@@ -1,2 +1,5 @@
 class APIError(ValueError):
-    pass
+    def __init__(self, message: str, status_code=0):
```
minghansun1: Status code prob shouldn't default to 0, otherwise this pr looks good

minghansun1 (approved):
LGTM

darrentweng (approved):

backend/utils/errors.py
esinx: thought about it maybe should just do `int | None`

